Issue #2, opened 2018-12-09
ImportError: cannot import name Directive with Sphinx 1.8.2


onkwon:
Here is the error message:

```
Traceback (most recent call last):
  File "cautodoc.py", line 20, in <module>
    from sphinx.util.compat import Directive
ImportError: cannot import name Directive
```

And I found [the issue](https://github.com/spinus/sphinxcontrib-images/issues/41) on the Sphinx Github issues list and removed

`from sphinx.util.compat import Directive`

line and replaced 

`from docutils.parsers.rst import directives`

with

`from docutils.parsers.rst import Directive, directives`

 in `cautodoc.py` fixed the problem.

jnikula:
Thanks for your report! Fixed by commit 92617df84ce3 ("cautodoc: fix dependency on deprecated module").

PR #3, opened 2018-12-09
cautodoc: fix Directive import error, #2

onkwon:


jnikula:
Thanks for your contribution! Incidentally, I received an almost identical fix just a few days ago, and pushed that on a first come, first served basis. The fix is 92617df84ce3 ("cautodoc: fix dependency on deprecated module").

PR #7, opened 2019-04-22
hawkmoth/parser: Display parser warnings and errors

mv0:
This patch prints out warnings, errors and fatal messages from clang
python bindings. Allows to trouble shooting to happen much faster
(i.e. missing include directories when resolving symbols)

Signed-off-by: Marius Vlad <marius.vlad@collabora.com>

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -19,6 +19,7 @@ from docutils.parsers.rst import directives, Directive
 from docutils.statemachine import ViewList
 from sphinx.util.nodes import nested_parse_with_titles
 from sphinx.util.docutils import switch_source_input
+from sphinx.util import logging
 
 from hawkmoth.parser import parse
 
@@ -30,6 +31,7 @@ class CAutoDocDirective(Directive):
     """Extract all documentation comments from the specified file"""
     required_argument = 1
     optional_arguments = 1
+    logger = logging.getLogger(__name__)
 
     # Allow passing a variable number of file patterns as arguments
     final_argument_whitespace = True
@@ -65,16 +67,16 @@ class CAutoDocDirective(Directive):
             filenames = glob.glob(env.config.cautodoc_root + '/' + pattern)
             if len(filenames) == 0:
                 fmt = 'Pattern "{pat}" does not match any files.'
-                env.app.warn(fmt.format(pat=pattern),
-                             location=(env.docname, self.lineno))
+                self.logger.warning(fmt.format(pat=pattern),
+                                    location=(env.docname, self.lineno))
                 continue
 
             for filename in filenames:
                 mode = os.stat(filename).st_mode
                 if stat.S_ISDIR(mode):
                     fmt = 'Path "{name}" matching pattern "{pat}" is a directory.'
-                    env.app.warn(fmt.format(name=filename, pat=pattern),
-                                 location=(env.docname, self.lineno))
+                    self.logger.warning(fmt.format(name=filename, pat=pattern),
+                                        location=(env.docname, self.lineno))
                     continue
 
                 # Tell Sphinx about the dependency and parse the file
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -42,13 +42,29 @@ class CAutoDocDirective(Directive):
     }
     has_content = False
 
+    def __display_parser_diagnostics(self, errors):
+        env = self.state.document.settings.env
+        for diag in errors:
+            severity, filename = diag[0], diag[1]
+            lineno, reason = diag[2], diag[3]
+            toprint = '{}:{}, {}'.format(filename, lineno, reason)
+            if severity == 2:
+                self.logger.warning(toprint, location(env.docname, self.lineno))
+            elif severity == 3:
+                self.logger.critical(toprint, location(env.docname, self.lineno))
+            else:
+                self.logger.error(toprint, location=(env.docname, self.lineno))
+
     def __parse(self, viewlist, filename):
         env = self.state.document.settings.env
 
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
 
-        comments = parse(filename, compat=compat, clang=clang)
+        comments, errors = parse(filename, compat=compat, clang=clang)
+        # at least -v
+        if errors and env.app.verbosity:
+            self.__display_parser_diagnostics(errors)
 
         for (comment, meta) in comments:
             lineoffset = meta['line'] - 1
--- a/hawkmoth/__main__.py
+++ b/hawkmoth/__main__.py
@@ -31,8 +31,13 @@ def main():
                         help='Verbose output.')
     args = parser.parse_args()
 
-    comments = parse_to_string(args.file, args.verbose,
-                               compat=args.compat, clang=args.clang)
+    comments, errors = parse_to_string(args.file, args.verbose,
+                                        compat=args.compat, clang=args.clang)
+    if errors:
+        for diag in errors:
+            severity, filename = diag[0], diag[1]
+            lineno, reason = diag[2], diag[3]
+            sys.stderr.write("{}:{}, {}\n".format(filename, lineno, reason))
     sys.stdout.write(comments)
 
 main()
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -34,7 +34,7 @@ Otherwise, documentation comments are passed through verbatim.
 """
 
 import itertools
-import sys
+import sys, re
 
 from clang.cindex import CursorKind
 from clang.cindex import Index, TranslationUnit
@@ -239,10 +239,29 @@ def _recursive_parse(comments, cursor, nest, compat):
 
     return [(doc, meta)]
 
+def extract_filename_line(source_loc):
+    valid = re.compile(r".*file\s'(.*)',\sline\s(\d+).*")
+    match = valid.match(source_loc)
+    filename = ''
+    line = -1
+    if match:
+        filename = match.group(1)
+        line = int(match.group(2))
+
+    return filename, line
+
+def parse_gather_diagnostics(severity, errors, diagnostics):
+    for diag in diagnostics:
+        if diag.severity >= severity:
+            filename, line = extract_filename_line(str(diag.location))
+            errors.extend([(diag.severity, filename, line, diag.spelling)])
+
 # return a list of (comment, metadata) tuples
 # options - dictionary with directive options
 def parse(filename, **options):
 
+    parser_errors = []
+    severity_diag = 2
     args = options.get('clang')
     if args is not None:
         args = [s.strip() for s in args.split(',') if len(s.strip()) > 0]
@@ -254,6 +273,8 @@ def parse(filename, **options):
     tu = index.parse(filename, args=args, options=
                      TranslationUnit.PARSE_DETAILED_PROCESSING_RECORD |
                      TranslationUnit.PARSE_SKIP_FUNCTION_BODIES)
+
+    parse_gather_diagnostics(severity_diag, parser_errors, tu.diagnostics)
 
     top_level_comments, comments = comment_extract(tu)
 
@@ -270,13 +291,13 @@ def parse(filename, **options):
     # Sort all elements by order of appearance.
     result.sort(key=lambda r: r[1]['line'])
 
-    return result
+    return result, parser_errors
 
 def parse_to_string(filename, verbose, **options):
     s = ''
-    comments = parse(filename, **options)
+    comments, parser_errors = parse(filename, **options)
     for (comment, meta) in comments:
         if verbose:
             s += ('# ' + str(meta) + '\n')
         s += comment + '\n'
-    return s
+    return s, parser_errors
--- a/test/test_hawkmoth.py
+++ b/test/test_hawkmoth.py
@@ -9,7 +9,9 @@ import testenv
 from hawkmoth.parser import parse_to_string
 
 def _get_output(input_filename, **options):
-    return parse_to_string(input_filename, False, **options)
+    output, errors = parse_to_string(input_filename, False, **options)
+    # FIXME: should handle errors
+    return output
 
 def _get_expected(input_filename, **options):
     return testenv.read_file(input_filename, ext='rst')


jnikula:
Thanks for the contribution! I think we'll want the ability to display the parser warnings and errors.

However, if you look at parser.py and __init__.py, you'll find that they depend on Clang xor Sphinx. This is to keep the design clean, and I'm not willing to give up on that, not yet at least. I think we need to look for a better way to report the messages.

mv0:
Hi,

Alright, but besides using plain print I don't see how exactly? Any suggestion would be nice! :-).

BrunoMSantos:
I think we should pass the messages together with relevant metadata (file and line number) in a Sphinx agnostic way back to the directive as part of the results. My gut feeling is that this shouldn't be too messy or hard. Do you see any issues?

mv0:
I've updated to use docstr to pass parser diagnostics. Mind taking another look? 

I have to admit this will be rather strange to see this kind of warnings in the documentation,
but at least it lets the user that something is clearly wrong.

BrunoMSantos:
Sorry, I can see the confusion, but this was not what I meant :/

I meant passing something like `return results, errors`. And then in `__init__.py` having a for loop where we print out the errors with the appropriate Sphinx interfaces while the results are added to the documentation. Indeed putting the errors in the documentation is weird and would lead to them being missed altogether in many cases.

BrunoMSantos:
I just submitted a pull request with my take on it. It's not functional... explanation will follow.

Sorry if I'm doing this wrong, I've never used this pull request thing in my life. Email FTW!

mv0:
Using this new version when running tests:

```
  self.source = open(source_path, mode, **kwargs)
.................x..................S 3, <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-50-struct.c', line 9, column 2>, unknown type name 'list_data_t'
.S 2, <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-70-function.c', line 9, column 17>, declaration of 'struct list' will not be visible outside of this function
S 2, <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-70-function.c', line 9, column 34>, declaration of 'enum mode' will not be visible outside of this function
.S 2, <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-80-compat.c', line 9, column 18>, declaration of 'struct list' will not be visible outside of this function
S 2, <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-80-compat.c', line 9, column 35>, declaration of 'enum mode' will not be visible outside of this function
....x.......
----------------------------------------------------------------------
Ran 50 tests in 1.777s

OK (expected failures=2)
````

**Comment by BrunoMSantos** [comment]
Yes, the printing method is wrong, but I was trying to mimic what is done a few lines below already. It's the idea of passing these things in 2 different groups and printing errors in `__init__.py` that I wanted to point to.

Thanks for the link, I think this is indeed the right way to register warnings and errors with Sphinx. I'll push something a bit more polished and with a few other errors I have in mind when I get the spare time.

Another issue is how we unit test the warnings and errors and adding relevant test cases.

File: hawkmoth/__init__.py [hunk 1]
```diff
@@ -48,7 +48,10 @@ def __parse(self, viewlist, filename):
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
 
-        comments = parse(filename, compat=compat, clang=clang)
+        comments, parser_errs = parse(filename, compat=compat, clang=clang)
```
BrunoMSantos: Minor thing: I think `errors` looks nicer. Also not a fan of `err`.

hawkmoth/parser.py
```diff
@@ -255,6 +256,11 @@ def parse(filename, **options):
                      TranslationUnit.PARSE_DETAILED_PROCESSING_RECORD |
                      TranslationUnit.PARSE_SKIP_FUNCTION_BODIES)
 
+    for diag in tu.diagnostics:
+        if diag.severity >= 2:
+            parser_errs.append('S {}, {}, {}'.format(diag.severity,
+                                                     diag.location,
+                                                     diag.spelling))
```
BrunoMSantos: Don't like this so much. I prefer to pass the information as a tuple as shown in my example. That way we can print different severity messages as warnings or errors. Note that this module should stand on its own, maybe we want to use the error stuff differently somewhere else without regexing a string.

On another note, I'd prefer `diag.location` to be split in `.file.name` and `.line`. That's more general and consistent once we add other non Clang related warnings and errors.

test/test_hawkmoth.py
```diff
@@ -9,7 +9,11 @@
 from hawkmoth.parser import parse_to_string
 
 def _get_output(input_filename, **options):
-    return parse_to_string(input_filename, False, **options)
+    output, errs = parse_to_string(input_filename, False, **options)
+    if errs:
+        for err in errs:
+            print(err)
```
BrunoMSantos: This is wrong, we should not print any of this here.

We need to create test cases to automatically check that the warnings and errors are generated appropriately and provide unit tests designed to produce them. I didn't do this in my example either by the way, but I believe it's needed. It's also a bit more work.

I'd hold off that work though until we settle the issue around the level of errors so that we don't waste time on half solutions.

hawkmoth/parser.py
```diff
@@ -255,6 +256,11 @@ def parse(filename, **options):
                      TranslationUnit.PARSE_DETAILED_PROCESSING_RECORD |
                      TranslationUnit.PARSE_SKIP_FUNCTION_BODIES)
 
+    for diag in tu.diagnostics:
+        if diag.severity >= 2:
```
BrunoMSantos: So I've tried it out and a missing header file is indeed level 4 (the highest), yet it does *not* affect the documentation. Clang does the right thing as far as I can see and just keeps going.

Maybe I didn't understand what you meant before? Or do you have an example where that occurs?

Strangely though, I added blatant syntax errors and this thing picks nothing up and documentation is still fine as long as the syntax errors don't involve blocks boundaries. It seems whatever Clang issues here is totally useless for our purposes since it doesn't matter for our task: the documentation.

I expect Clang collects those other errors in some way. I think we need to figure out how, because those are the errors that actually matter for hawkmoth. Would you agree?

hawkmoth/__init__.py
```diff
@@ -65,16 +70,16 @@ def run(self):
             filenames = glob.glob(env.config.cautodoc_root + '/' + pattern)
             if len(filenames) == 0:
                 fmt = 'Pattern "{pat}" does not match any files.'
-                env.app.warn(fmt.format(pat=pattern),
-                             location=(env.docname, self.lineno))
+                self.logger.warning(fmt.format(pat=pattern),
+                               location=(env.docname, self.lineno))
```
BrunoMSantos: Bad indentation. Same below.

hawkmoth/__main__.py
```diff
@@ -31,8 +31,11 @@ def main():
                         help='Verbose output.')
     args = parser.parse_args()
 
-    comments = parse_to_string(args.file, args.verbose,
-                               compat=args.compat, clang=args.clang)
+    comments, parser_errs = parse_to_string(args.file, args.verbose,
+                                            compat=args.compat, clang=args.clang)
+    if parser_errs:
+        for err in parser_errs:
+            print(err)
```
BrunoMSantos: Sphinx prints everything on stdout unfortunately, but I don't think we need to follow that here. Maybe stderr would be better here? Don't think it matters much though.

hawkmoth/parser.py
mv0: > Maybe I didn't understand what you meant before? Or do you have an example where that occurs?

In our case, the parser couldn't find the proper header files such that macros weren't expanded leading to the comment being added but not the function definition. I debugged a bit of it by creating a simple python script that that iterated over all the symbols (it literally stopped when it couldn't find the symbol). 

And for an example:

Here's what Hawmoth generates with the function_attributes branch for test/example-72-function-attrs-with-macros.c file:

```

.. c:function:: int hidden_frob(struct list * list, enum mode mode)

   List frobnicator with visibility hidden attribute.

   :param list: The list to frob.
   :param mode: The frobnication mode.
   :return: 0 on success, non-zero error code on error.
   :since: v0.1

**Visibility** hidden
```

I would then remove the macros and place them into a header file (and include that header file, but without a proper path):

Here's what it prints (just the comments):

```

List frobnicator with visibility hidden attribute.

:param list: The list to frob.
:param mode: The frobnication mode.
:return: 0 on success, non-zero error code on error.
:since: v0.1

```

>  I expect Clang collects those other errors in some way. I think we need to figure out how, because those are the errors that actually matter for hawkmoth. Would you agree?

Yes I agree. The reason which you do not see that is because of `TranslationUnit.PARSE_SKIP_FUNCTION_BODIES` passed to parse method. Once you remove that you'll also see those errors.  @jnikula do you think we should remove this such that we display compiler errors? 

I personally don't have strong opinions here. Somehow not being able to build your code takes bigger precedence than the ability to generate documentation out of it so I'd expect that code builds properly and then you generate documentation out of it. I'd argue that tools that generate documentation should take care to make sure the code is valid. IDK you decide.

hawkmoth/__init__.py
mv0: Okay, will fix.

hawkmoth/__main__.py
mv0: Will use stderr then.

test/test_hawkmoth.py
mv0: So warnings are like:

```
2 <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-70-function.c', line 9, column 17> declaration of 'struct list' will not be visible outside of this function
```

While errors are like:
```
S 3, <SourceLocation file '/home/mvlad/src/hawkmoth/test/example-50-struct.c', line 9, column 2>, unknown type name 'list_data_t'
```

And fatal are when files are not found at all.
```
S 4, <SourceLocation file '/tmp/bla.c', line 1, column 10>, 'nofile.h' file not found
```

I haven't touched test part at all so IDK how this should behave but tests like these are quite easy to add.

hawkmoth/parser.py
jnikula: I regret I haven't been able to look into this in depth.

For things to work properly in a complex project I expect you have to pass the CFLAGS to Clang through `cautodoc_clang` config option or `clang` directive option. That's not about the project building in general, it's about running Clang with the same flags.

The question is, if you fail to do so, and you get incomplete/incorrect documentation, what are your options? I'm inclined to think we should keep the Clang build warnings to a minimum in the happy day scenario. This would mean using the Sphinx logging API, with suitable loglevels to keep the happy day scenario clean. The user can then use, say, `sphinx-build -vvv` or some number of `-v`s to get more data.

`TranslationUnit.PARSE_SKIP_FUNCTION_BODIES` is an optimization. We don't usually need anything from within the functions. If we ensure we don't rely on the state of that option in the parser, we could leave it out depending on Sphinx loglevel to gain more debug info.

mv0: > For things to work properly in a complex project I expect you have to pass the CFLAGS to Clang through cautodoc_clang config option or clang directive option. That's not about the project building in general, it's about running Clang with the same flags.

Even with those properly added standard include files require standard clang header files (which are _not_ added by the parser). One thing that would like would be to improve hawkmoth documentation to mention this somewhere because I struggled a bit with it. 

> The question is, if you fail to do so, and you get incomplete/incorrect documentation, what are your options? I'm inclined to think we should keep the Clang build warnings to a minimum in the happy day scenario. This would mean using the Sphinx logging API, with suitable loglevels to keep the happy day scenario clean. The user can then use, say, sphinx-build -vvv or some number of -vs to get more data.

Yep, I tend to agree with this, and it seems reasonable to me.

BrunoMSantos: Ok, so I must admit you've both turned my own argument against me, and quite rightly so!

I was hoping for it to go the other way around: happy day scenario means not caring about missing stuff (including half the flags), but caring about syntax that may break the parser (e.g. bad function boundaries affecting placement of comments). Turns out both can break documentation, so my argument is out.

So yeah, I guess it does make sense to assume all flags are passed to Clang. And as long as we assume flags are being passed entirely, then level 2 seems appropriate.

We can still argue in favour of removing `TranslationUnit.PARSE_SKIP_FUNCTION_BODIES`, but it can come separately after we gain more experience with this.

test/test_hawkmoth.py
BrunoMSantos: Yup, and I'm kindly requesting they go in as part of this series. :)

Rather than polluting stdout with this, we must compare with expected warnings/errors and just say: it passed/failed. We also need tests for it.

On the presentation side of things, I'm not happy with it. You're using 'Clang speak', but we will have warnings and errors that have nothing to do with Clang so we must have a generic way of showing them.

1. Severity must be about registering warnings or errors, the user has no need for some random number with a random `S` preceding it. The user must get `WARNING:` or `ERROR:` or whatever Sphinx supports.
2. File name, line and column must be presented in a Clang and Python agnostic way. You're currently dumping the `__repr__` method of a Clang object or whatever that is (I made this comment somewhere else already).

This is partly why I'd like to see some other kind of warnings and errors being raised in this series already. It would help consolidate the idea in a more general way.

mv0: I understand that this has be improved, but in the same time, this will take much longer to do properly and it is orthogonal to current scope (imo this escalate to something else :)).

For the time being, I'm suggesting to improve testing facility in a new PR.

mv0:
So, I've updated this a bit. One thing which I've mentioned in @BrunoMSantos (now defunct) PR about the same thing. The logging facility from sphinx seem to display the same level (that is warning/critical/error all pre-pend the warning text with a `WARNING` message):

With the new changes the following will be printed (upon using `-v` on `sphinx-build`):

```
/home/mvlad/src/weston/libweston/doc/usage/compositor.rst:28: WARNING: /home/mvlad/src/weston//include/libweston/libweston.h:35, 'stdbool.h' file not found
/home/mvlad/src/weston/libweston/doc/usage/compositor.rst:42: WARNING: /home/mvlad/src/weston//libweston/compositor.c:29, 'config.h' file not found
/home/mvlad/src/weston//libweston/compositor.c:4819: WARNING: Inline emphasis start-string without end-string.
/home/mvlad/src/weston//libweston/compositor.c:4825: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/mvlad/src/weston//libweston/compositor.c:4829: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/mvlad/src/weston//libweston/compositor.c:4865: WARNING: Inline emphasis start-string without end-string.
/home/mvlad/src/weston//libweston/compositor.c:4871: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/mvlad/src/weston//libweston/compositor.c:4875: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/mvlad/src/weston/libweston/doc/usage/compositor.rst:50: WARNING: /home/mvlad/src/weston//include/libweston/windowed-output-api.h:33, 'libweston/plugin-registry.h' file not found
/home/mvlad/src/weston/libweston/doc/usage/input.rst:10: WARNING: /home/mvlad/src/weston//libweston/input.c:28, 'config.h' file not found
/home/mvlad/src/weston//libweston/input.c:198: WARNING: duplicate C object description of bool, other instance in /home/mvlad/src/weston/libweston/doc/usage/compositor.rst
/home/mvlad/src/weston/libweston/doc/usage/input.rst:16: WARNING: /home/mvlad/src/weston//libweston/input.c:28, 'config.h' file not found
/home/mvlad/src/weston/libweston/doc/usage/input.rst:22: WARNING: /home/mvlad/src/weston//libweston/input.c:28, 'config.h' file not found
/home/mvlad/src/weston/libweston/doc/usage/input.rst:28: WARNING: /home/mvlad/src/weston//libweston/input.c:28, 'config.h' file not found
```

Those `WARNING` about missing file are actually through: `sphinx.util.logging.error()` and not `sphinx.util.logging.warning()`.

test/test_hawkmoth.py
BrunoMSantos: If @jnikula's ok with that, so am I. The other comments still apply, but I think you addressed those at least in part.

BrunoMSantos:
> So, I've updated this a bit. One thing which I've mentioned in @BrunoMSantos (now defunct) PR about the same thing. The logging facility from sphinx seem to display the same level (that is warning/critical/error all pre-pend the warning text with a WARNING message):

Not ideal, but if that's the case, so be it. We won't touch Sphinx code, but we can still use the right functions :)

> With the new changes the following will be printed (upon using -v on sphinx-build):

Hm, maybe we're using the wrong interfaces then? I *think* other extensions print stuff directly in Sphinx's stdout output. And I do think this output includes `WARNING:` and `ERROR:` messages by the way.

I'll dig around a bit. Personally I haven't looked at the proper way of doing it but I don't think this should be hidden behind `-v` if possible. @jnikula, was this your expectation?

hawkmoth/parser.py
```diff
@@ -239,10 +239,29 @@ def _recursive_parse(comments, cursor, nest, compat):
...
+    if match:
+        filename = match.group(1)
+        line = int(match.group(2))
+
+    return filename, line
+
+def parse_gather_diagnostics(severity, errors, diagnostics):
+    for diag in diagnostics:
+        if diag.severity >= severity:
+            filename, line = extract_filename_line(str(diag.location))
```
BrunoMSantos: Use `diag.location.file.name` and `diag.location.line` (if I remember correctly). No need for importing `re` either.

hawkmoth/parser.py
```diff
@@ -239,10 +239,29 @@ def _recursive_parse(comments, cursor, nest, compat):
...
+            errors.extend([(diag.severity, filename, line, diag.spelling)])
+
 # return a list of (comment, metadata) tuples
 # options - dictionary with directive options
 def parse(filename, **options):
 
+    parser_errors = []
```
BrunoMSantos: Missed opportunity for rename :)

hawkmoth/parser.py
```diff
@@ -239,10 +239,29 @@ def _recursive_parse(comments, cursor, nest, compat):
...
+        filename = match.group(1)
+        line = int(match.group(2))
+
+    return filename, line
+
+def parse_gather_diagnostics(severity, errors, diagnostics):
+    for diag in diagnostics:
+        if diag.severity >= severity:
+            filename, line = extract_filename_line(str(diag.location))
+            errors.extend([(diag.severity, filename, line, diag.spelling)])
```
BrunoMSantos: Still don't think we should be passing Clang interfaces up the chain. What does 2 mean for us? We know it's a warning in Clang, but when we add other warnings and errors, are we also going to use 2, 3 and 4? Do we skip 0 and 1 entirely? Feels a bit weird...

I was thinking we could hide Clang's severity levels behind a `class ErrorLevel(Enum)` or something (look into `docstr.py` for an example) we could then leverage everywhere consistently.

test/test_hawkmoth.py
```diff
@@ -9,7 +9,9 @@
 from hawkmoth.parser import parse_to_string
 
 def _get_output(input_filename, **options):
-    return parse_to_string(input_filename, False, **options)
+    output, errors = parse_to_string(input_filename, False, **options)
+    # FIXME: should handle errors
+    return output
```
BrunoMSantos: Yup, that's ok for now if @jnikula agrees to postpone the testing side of things.

BrunoMSantos:
> Hm, maybe we're using the wrong interfaces then? I think other extensions print stuff directly in Sphinx's stdout output. And I do think this output includes WARNING: and ERROR: messages by the way.

Nah, looking at other seriously maintained extensions' code they all seem to do the same. Never mind that comment.

hawkmoth/__init__.py
```diff
@@ -42,13 +42,29 @@ class CAutoDocDirective(Directive):
...
+    def __display_parser_diagnostics(self, errors):
+        env = self.state.document.settings.env
+        for diag in errors:
+            severity, filename = diag[0], diag[1]
+            lineno, reason = diag[2], diag[3]
+            toprint = '{}:{}, {}'.format(filename, lineno, reason)
+            if severity == 2:
+                self.logger.warning(toprint, location(env.docname, self.lineno))
+            elif severity == 3:
+                self.logger.critical(toprint, location(env.docname, self.lineno))
```
BrunoMSantos: `location=`, same above.

hawkmoth/__init__.py
```diff
@@ -42,13 +42,29 @@ class CAutoDocDirective(Directive):
     }
     has_content = False
 
+    def __display_parser_diagnostics(self, errors):
+        env = self.state.document.settings.env
+        for diag in errors:
+            severity, filename = diag[0], diag[1]
+            lineno, reason = diag[2], diag[3]
+            toprint = '{}:{}, {}'.format(filename, lineno, reason)
```
BrunoMSantos: This repeats the filename and line number. Real example:
`/home/bruno/Documents/projects/bms-contrib/hawkmoth/doc/examples.rst:262: WARNING: /home/bruno/Documents/projects/bms-contrib/hawkmoth/doc/examples/example-70-function.c:9, declaration of 'struct list' will not be visible outside of this function`

This must be `self.logger.warning(reason, location(env.docname, self.lineno))` and so on below. You can get rid of this variable.

BrunoMSantos: Wait, actually it doesn't repeat it, there's just 2 sources involved of course. Sorry about that, I think it's probably best to show both.

BrunoMSantos:
So I fixed the issues I found in @mv0's series and added tests for it (no actually new test cases, I think existing ones are ok for now as explained in the commit message).

@mv0, can you double check to see if I didn't butcher it somehow? I fixed a few other tiny things I hadn't commented on for style consistency, but nothing big.

https://gitlab.com/bms-contrib/hawkmoth/tree/missing_diagnostics

By the way, I used to keep a mirror in GitHub, but I found that the automatic mirroring was kind of faulty and I don't want to maintain it manually so I got rid of it for the time being. Doesn't affect my usual workflow, but may hamper participation with GitHub folk. We'll see...

mv0:
> So I fixed the issues I found in @mv0's series and added tests for it (no actually new test cases, I think existing ones are ok for now as explained in the commit message).
> 
> @mv0, can you double check to see if I didn't butcher it somehow? I fixed a few other tiny things I hadn't commented on for style consistency, but nothing big.
> 
> https://gitlab.com/bms-contrib/hawkmoth/tree/missing_diagnostics
> 
Thanks for polishing them up!  Left a comment there about missing verbosity.

hawkmoth/__init__.py
mv0: Yep, I think it is better this way even though at first sight it looks strange.

hawkmoth/__init__.py
mv0: Yep, you already fixed this!

hawkmoth/parser.py
mv0: Yep, with that it looks much more cleaner in your version.

mv0:
Makes no sense to keep this open. @BrunoMSantos sent revised version over https://www.freelists.org/post/hawkmoth/PATCH-v2-13-hawkmoth-remove-parse-to-string
and 
https://www.freelists.org/post/hawkmoth/PATCH-04-Add-diagnostics-support-and-extend-test-framework-for-them

PR #8, opened 2019-04-22
hawkmoth: Display function attributes

mv0:
This adds some common attributes (like const, pure, and visibility).
Without this we have no way of knowing if the functions have such
attributes.

Signed-off-by: Marius Vlad <marius.vlad@collabora.com>

### DIFF
--- a/test/example-72-function-attrs-with-macros.rst
+++ b/test/example-72-function-attrs-with-macros.rst
@@ -1,0 +1,24 @@
+
+.. c:function:: int hidden_frob(struct list * list, enum mode mode)
+
+   List frobnicator with visibility hidden attribute.
+
+   :param list: The list to frob.
+   :param mode: The frobnication mode.
+   :return: 0 on success, non-zero error code on error.
+   :since: v0.1
+
+**Visibility** hidden
+
+
+.. c:function:: int default_frob(struct list * list, enum mode mode)
+
+   List frobnicator with visibility default attribute.
+
+   :param list: The list to frob.
+   :param mode: The frobnication mode.
+   :return: 0 on success, non-zero error code on error.
+   :since: v0.1
+
+**Visibility** default
+
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -89,14 +89,15 @@ def comment_extract(tu):
     return top_level_comments, comments
 
 def _result(comment, cursor=None, fmt=docstr.Type.TEXT, nest=0,
-            name=None, ttype=None, args=None, compat=None):
+            name=None, ttype=None, args=None, compat=None, attr=None):
 
     # FIXME: docstr.generate changes the number of lines in output. This impacts
     # the error reporting via meta['line']. Adjust meta to take this into
     # account.
 
     doc = docstr.generate(text=comment.spelling, fmt=fmt,
-                          name=name, ttype=ttype, args=args, transform=compat)
+                          name=name, ttype=ttype, args=args, transform=compat,
+                          attributes=attr)
 
     doc = docstr.nest(doc, nest)
 
@@ -105,6 +106,7 @@ def _result(comment, cursor=None, fmt=docstr.Type.TEXT, nest=0,
         meta['cursor.kind']        = cursor.kind,
         meta['cursor.displayname'] = cursor.displayname,
         meta['cursor.spelling']    = cursor.spelling
+        meta['cursor.attributes']  = attr
 
     return [(doc, meta)]
 
@@ -136,6 +138,25 @@ def _get_macro_args(cursor):
             break
 
     return None
+
+# add more function attributes here
+def parse_attributes(attributes, cursor):
+    entry = {}
+
+    if cursor.kind == CursorKind.VISIBILITY_ATTR:
+        entry['Attribute'] = 'Visibility'
+    elif cursor.kind == CursorKind.PURE_ATTR:
+        entry['Attribute'] = 'Pure'
+    elif cursor.kind == CursorKind.CONST_ATTR:
+        entry['Attribute'] = 'Const'
+    elif cursor.kind == CursorKind.ANNOTATE_ATTR:
+        entry['Attribute'] = 'Annonate'
+
+    if 'Attribute' in entry and cursor.spelling:
+        entry['Value'] = cursor.spelling
+
+    if 'Attribute' in entry:
+        attributes.append(entry)
 
 def _recursive_parse(comments, cursor, nest, compat):
     comment = comments[cursor.hash]
@@ -205,10 +226,12 @@ def _recursive_parse(comments, cursor, nest, compat):
         # FIXME: children may contain extra stuff if the return type is a
         # typedef, for example
         args = []
+        attributes = []
         for c in cursor.get_children():
             if c.kind == CursorKind.PARM_DECL:
                 args.append('{ttype} {arg}'.format(ttype=c.type.spelling,
                                                    arg=c.spelling))
+            parse_attributes(attributes, c)
 
         if cursor.type.is_function_variadic():
             args.append('...')
@@ -217,7 +240,8 @@ def _recursive_parse(comments, cursor, nest, compat):
         ttype = cursor.result_type.spelling
 
         return _result(comment, cursor=cursor, fmt=fmt, nest=nest,
-                       name=name, ttype=ttype, args=args, compat=compat)
+                       name=name, ttype=ttype, args=args, compat=compat,
+                       attr=attributes)
 
     # FIXME: If we reach here, nothing matched. This is a warning or even error
     # and it should be logged, but it should also return an empty list so that
--- a/hawkmoth/util/docstr.py
+++ b/hawkmoth/util/docstr.py
@@ -35,7 +35,7 @@ _doc_fmt = {
     Type.MEMBER:     (1, '\n.. c:member:: {ttype} {name}\n\n{text}\n'),
     Type.MACRO:      (1, '\n.. c:macro:: {name}\n\n{text}\n'),
     Type.MACRO_FUNC: (1, '\n.. c:function:: {name}({args})\n\n{text}\n'),
-    Type.FUNC:       (1, '\n.. c:function:: {ttype} {name}({args})\n\n{text}\n')
+    Type.FUNC:       (1, '\n.. c:function:: {ttype} {name}({args})\n\n{text}\n{attr}')
 }
 
 def _strip(comment):
@@ -67,8 +67,22 @@ def nest(text, nest):
     """
     return re.sub('(?m)^(?!$)', '   ' * nest, text)
 
+# injects reST
+def generate_attributes(attributes):
+    attrs = ''
+    if attributes is not None:
+        for attr in attributes:
+            k = attr['Attribute']
+            if k:
+                # not all attributes have a value
+                if 'Value' in attr:
+                    attrs += '\n**{}** {}\n'.format(k, attr['Value'])
+                else:
+                    attrs += '\n**{}**\n'.format(k)
+    return attrs
+
 def generate(text, fmt=Type.TEXT, name=None,
-             ttype=None, args=None, transform=None):
+             ttype=None, args=None, transform=None, attributes=None):
     """
     Generate reST documentation string.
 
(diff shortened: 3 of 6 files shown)


jnikula:
Note to self, I should really write some guidelines for contributing. One of requirements pretty much is that all functionality that impacts the output must include a test case. See the test/ subdirectory. (And note to self #2, that subdirectory could use a small README...)

Anyway, it would really be much easier to see what the proposed change does and what the output is with a test case.

Here, I'm not immediately convinced by looking at the code alone. In pure C, what do visibility, pure, and annotate mean in practice? In the output, shouldn't const rather be part of the function signature?

mv0:
Hi,

Adding a test is something I though on but was ultimately too lazy to add it. But I will if you believe this has the chance to make it in.

Regarding the attributes themselves: 
`const` in return type is something differently from `__attribute__`. The output does include `const` as return type but in case of `__attributes__` this is ignored entirely by hawkmoth. 

>visibility ("visibility_type")
This attribute affects the linkage of the declaration to which it is attached. It can be applied to variables (see Common Variable Attributes) and types (see Common Type Attributes) as well as functions.
There are four supported visibility_type values: default, hidden, protected or internal visibility.

For instance we compile everything with visibility hidden when we use default on the symbols we want to export. 

I've `const`, `pure`, `annotate` as this might seem useful to others (and `annotate` seems only clang related)

> const
For example, `int square (int) __attribute__ ((const));`
tells GCC that subsequent calls to function square with the same argument value can be replaced by the result of the first call regardless of the statements in between.

> pure
Calls to functions that have no observable effects on the state of the program other than to return a value may lend themselves to optimizations such as common subexpression elimination. Declaring such functions with the pure attribute allows GCC to avoid emitting some calls in repeated invocations of the function with the same argument values.

More here: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

BrunoMSantos:
Interesting idea, but in general I think this mostly clutters the documentation with very little to no benefit. If the attributes somehow have an important role in how the function is to be used, then that probably should be written down in prose rather than letting the reader infer it from criptic `__attribute__` no less.

And let's not forget that attributes are often hidden behind macros that may have totally different names and even combine many attributes (e.g. `__init` in the kernel to name one). What happens then? Seeing `__init` there wouldn't be particularly useful, but the resolved attributes carry no meaning at all (go and check for yourself the huge list of attributes that simple `__init` expands to... not pretty and not useful).

If you can convince @jnikula to merge this, then I'd like 2 conditions to be met:

1. Macros hiding the attributes are not expanded. We need test cases to evaluate this behaviour on top of the more normal ones.
Unfortunately I fear this may invalidate the whole concept as it stands, but I haven't tested it nor seen how Clang's AST captures such cases.

2.  I'd like it hidden behind an off by default option (`:attributes:` perhaps?). Documentation clutter should be given no incentive, let alone made mandatory.
Options should also be added sparingly though as they add maintenance cost, hence my reluctance in voting this up.

mv0:
> Interesting idea, but in general I think this mostly clutters the documentation with very little to no benefit.

Well, in our case, it would be quite useful to have this printed somewhere close to the symbol. Like I've said we build a few libraries and some application on top of those libraries. Symbols are only visible if we explicitly mark them as such. Users wanting to use the library to build their own custom compositors would need to know which symbols are available. Parallel to this, the ability to group by function attributes is something we pondered a few times, but I don't see how to achieve this given that only Hawkmoth knows about AST while Sphinx has not clue about it. So the ability to group by function attributes is something that you'll probably reject from the start.  

> If the attributes somehow have an important role in how the function is to be used, then that probably should be written down in prose rather than letting the reader infer it from criptic __attribute__ no less.

It is not criptic at all, you need a way to differentiate compiler hints than the language itself. Well, on the hand in the case of grouping of symbols you would rather have it not touch the source code, yet here you would. I personally don't have a strong opinion, but I see it as a superfluous way of doing things (as you already mark this symbol with an attribute).

> And let's not forget that attributes are often hidden behind macros that may have totally different names and even combine many attributes (e.g. __init in the kernel to name one). What happens then?

I've used Hawkmoth on some of our code, the C-preprocessor is ran before creating the AST so the macros are expanded when Hawkmoth iterates over cursors it sees the function attribute as a different node. In our case macros are in installed standard header files, in other words, the macros itself are not even part of the project source code. 

> Macros hiding the attributes are not expanded. We need test cases to evaluate this behaviour on top of the more normal ones.
Unfortunately I fear this may invalidate the whole concept as it stands, but I haven't tested it nor seen how Clang's AST captures such cases.

I'm in the process of adding further tests on this, also including ones that hide the `__attribute__ behind a macro. 

> I'd like it hidden behind an off by default option (:attributes: perhaps?). Documentation clutter should be given no incentive, let alone made mandatory.
Options should also be added sparingly though as they add maintenance cost, hence my reluctance in voting this up.

Alright, so I suppose something like would printing of function/symbol attributes? (sorry but it seems that code doc messes the formatting a bit)

`
 .. c:autodoc:: path/to/file.c
   :attributes:
`

jnikula:
> Parallel to this, the ability to group by function attributes is something we pondered a few times, but I don't see how to achieve this given that only Hawkmoth knows about AST while Sphinx has not clue about it. So the ability to group by function attributes is something that you'll probably reject from the start.

I was trying to say that I'd prefer it if we didn't have code that has intimate knowledge about both Sphinx *and* Clang. The current meta information passed from the parser to the Sphinx extension is already a bit leaky, but the intention is that's only used for debugging/diagnostics.

This doesn't prevent us from requesting information or grouping or sorting from the Sphinx side and passing it on to the Clang side, and back. I just want that interface to be relatively clean from both Sphinx and Clang internals.

jnikula:
> EDIT: Sorry about the formatting. Replying directly from email is apparently not a good idea. This would be well moved to the mailing list by the way...

I think feature requests as github issues would work fine too. More from the perspective "we have functions with and without certain attributes, and we need the ability to highlight or group them differently in the documentation".

mv0:
> This doesn't prevent us from requesting information or grouping or sorting from the Sphinx side and passing it on to the Clang side, and back. I just want that interface to be relatively clean from both Sphinx and Clang internals.

Alright, this seems reasonable to me.

> Nice! Is this an open source project by any chance? It would be
interesting to have a look at it. There are no known users so far.

Yep: https://gitlab.freedesktop.org/wayland/weston

I've updated this series with some test examples. I'll add that `:attribute:` keyword next.

BrunoMSantos:
> Yep: https://gitlab.freedesktop.org/wayland/weston

Great to know, thanks. Let us know once you have a branch using hawkmoth ;)

Thanks for the updates!

test/example-72-function-attrs-with-macros.rst
```diff
@@ -0,0 +1,24 @@
...
+.. c:function:: int hidden_frob(struct list * list, enum mode mode)
+
+   List frobnicator with visibility hidden attribute.
+
+   :param list: The list to frob.
+   :param mode: The frobnication mode.
+   :return: 0 on success, non-zero error code on error.
+   :since: v0.1
+
+**Visibility** hidden
```
BrunoMSantos: I haven't checked the compiled output, but assuming this looks fine in the compiled documentation I have only one other comment: I'd like to see `**EXPORT_HIDEN**` instead and then you can document the macro itself if you want. That was the whole point I tried to make with the `__init` macro. Note that the macro also carries a lot more meaning in the context of a such a project in my opinion. Maybe I'm alone on this one though...

A more interesting question is whether that would mean that we should also get `__attribute__((pure))` verbatim in the other example. I know it's uglier, but I'm trying to find common ground rather than hard coding things that are specific to how one project uses them. And the common ground is to do as little as possible.

Note I haven't looked at Clang's AST for this and also, what would Clang put there if the define was elsewhere and not reachable when generating the docs? I believe Clang won't include other header files in the way we use it, but I haven't touched this in a couple of months, I may be forgetting stuff.

mv0: > I have only one other comment: I'd like to see **EXPORT_HIDEN** instead and then you can document the macro itself if you want. 

Placing the macro name directly means to extract it from the tokens comments. As the C-preprocessor is ran it will replace the macro with the attribute itself so when the parser goes over the AST it won't be present. I for one would've like to see it present as well, though, in this context, putting in front of the symbol wouldn't do any justice (as it isn't a return type). Doing this would probably violate one of the design principles of hawkmoth?

> I haven't checked the compiled output, but assuming this looks fine in the compiled documentation

It looks pretty OK to me: https://people.collabora.com/~mvlad/libweston-hawkmoth-group/usage/input.html#this-displays-only-functions-in-group-seat

> A more interesting question is whether that would mean that we should also get __attribute__((pure)) verbatim in the other example. I know it's uglier, but I'm trying to find common ground rather than hard coding things that are specific to how one project uses them. And the common ground is to do as little as possible.

In my "experiments" not all attributes are "equal". Some like (`const`, `pure`, `visibility`) are tagged as a node in the AST. Others like `noreturn` are in the function body but not as return types: 

```
`-FunctionDecl 0x350d4f0 <line:3:18, line:36:51> col:6 no_ret_frob 'void (struct list *, enum mode) __attribute__((noreturn))'
  |-ParmVarDecl 0x350d1e0 <col:18, col:31> col:31 list 'struct list *'
  |-ParmVarDecl 0x350d358 <col:37, col:47> col:47 mode 'enum mode':'enum mode'
  `-FullComment 0x350da20 <line:28:3, line:33:15>
    |-ParagraphComment 0x350d930 <line:28:3, col:44>
    | `-TextComment 0x350d900 <col:3, col:44> Text=" List frobnicator with noreturn attribute."
    `-ParagraphComment 0x350d9f0 <line:30:3, line:33:15>
      |-TextComment 0x350d950 <line:30:3, col:33> Text=" :param list: The list to frob."
      |-TextComment 0x350d970 <line:31:3, col:38> Text=" :param mode: The frobnication mode."
      |-TextComment 0x350d990 <line:32:3, col:55> Text=" :return: 0 on success, non-zero error code on error."
      `-TextComment 0x350d9b0 <line:33:3, col:15> Text=" :since: v0.1"
```

So to get back to your questions: you won't see `pure` attribute in the body just in the node as Pure attribute:

```
|-FunctionDecl 0x256e328 <test/example-71-function-attrs.c:9:1, line:10:48> col:5 pure_frob 'int (struct list *, enum mode)'
| |-ParmVarDecl 0x256e090 <col:15, col:28> col:28 list 'struct list *'
| |-ParmVarDecl 0x256e208 <col:34, col:44> col:44 mode 'enum mode':'enum mode'
| |-PureAttr 0x256e3d8 <line:9:16>
| `-FullComment 0x2596d50 <line:2:3, line:7:15>
|   |-ParagraphComment 0x2596c60 <line:2:3, col:40>
|   | `-TextComment 0x2596c30 <col:3, col:40> Text=" List frobnicator with pure attribute."
|   `-ParagraphComment 0x2596d20 <line:4:3, line:7:15>
|     |-TextComment 0x2596c80 <line:4:3, col:33> Text=" :param list: The list to frob."
|     |-TextComment 0x2596ca0 <line:5:3, col:38> Text=" :param mode: The frobnication mode."
|     |-TextComment 0x2596cc0 <line:6:3, col:55> Text=" :return: 0 on success, non-zero error code on error."
|     `-TextComment 0x2596ce0 <line:7:3, col:15> Text=" :since: v0.1"
```

mv0: ping @BrunoMSantos would be nice to get some feedback here as well.

BrunoMSantos: Sorry, day job gets in the way, I'm planning some more serious follow ups over the weekend :)

Yeah, this one's a bugger. As a user I actually don't care because I would never add this to the documentation myself (hence my request for this being optional), but as a developer interested in making it work for everyone who may want it, I still think my points are valid.

All I can say is that your answer is very detailed and raises interesting points, but it only tells me this can't be done right --- i.e. simple/robust and universal. I'd be willing to sacrifice simplicity if we achieved universality, and I mean that in 2 ways: in the sense that it handles any attribute or macro the same way and in that it makes no assumptions about what they might mean, right down to the expansion of macros (`__init` reads totally different from say 'place this code in cold section, optimize for space, etc.'). If you can find a way of doing that, I may be willing to support the idea myself.

I also think that some of your goals in these PRs are interdependent and currently biased towards your existing Doxygen documentation. Have you considered if any of the suggested solutions would work for you? I.e. splitting API into separate headers (or separate it with new directives or whatever comes out of the grouping PR) and / or documenting in plain English if a certain function is part of the API or not.

If not, I think it's best if I back down and let the dictator (@jnikula) decide. I really have no further input at the design level :/

mv0: > I also think that some of your goals in these PRs are interdependent and currently biased towards your existing Doxygen documentation.

jnikula:
The request has gone stale, and I think the conclusion was that this is a more advanced feature than we like to implement. (At least for the time being.)

PR #9, opened 2019-04-22
Group filtering

mv0:
This series add 'ingroup' option, similar to doxygen which allows grouping/filtering of symbols.

### DIFF
--- a/doc/extension.rst
+++ b/doc/extension.rst
@@ -55,6 +55,23 @@ The extension has a few configuration options that can be set in ``conf.py``:
    source, typically to define macros for conditional compilation, for example
    ``-DHAWKMOTH``. No arguments are passed by default.
 
+.. py:data:: cautodoc_ingroup
+
+   Filters the symbols hawkmoth will display. Use this to limit the ammount of
+   symbols hawkmoth will generate.
+
+   In your comment block add the following:
+
+   .. code-block:: python
+      @ingroup group_name
+
+
+   While in reST the following will only print those symbols tagged in comment
+   block:
+
+   .. code-block:: python
+      :ingroup: group_name
+
 Directive
 ---------
 
@@ -70,6 +87,9 @@ This module provides the following new directive:
    option.
 
    The ``clang`` option overrides the :data:`cautodoc_clang` configuration
+   option.
+
+   The ``ingroup`` option overrides the :data:`cautodoc_ingroup` configuration
    option.
 
 Examples
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -38,6 +38,7 @@ class CAutoDocDirective(Directive):
     option_spec = {
         'compat': directives.unchanged_required,
         'clang': directives.unchanged_required,
+        'ingroup': directives.unchanged_required,
     }
     has_content = False
 
@@ -46,8 +47,9 @@ class CAutoDocDirective(Directive):
 
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
+        ingroup = self.options.get('ingroup', env.config.cautodoc_ingroup)
 
-        comments = parse(filename, compat=compat, clang=clang)
+        comments = parse(filename, compat=compat, clang=clang, ingroup=ingroup)
 
         for (comment, meta) in comments:
             lineoffset = meta['line'] - 1
@@ -94,6 +96,7 @@ def setup(app):
     app.add_config_value('cautodoc_root', app.confdir, 'env')
     app.add_config_value('cautodoc_compat', None, 'env')
     app.add_config_value('cautodoc_clang', None, 'env')
+    app.add_config_value('cautodoc_ingroup', None, 'env')
     app.add_directive_to_domain('c', 'autodoc', CAutoDocDirective)
 
     return dict(version = __version__,
--- a/hawkmoth/__main__.py
+++ b/hawkmoth/__main__.py
@@ -27,12 +27,14 @@ def main():
                         help='Compatibility options. See cautodoc_compat.')
     parser.add_argument('--clang', metavar='PARAM[,PARAM,...]',
                         help='Arguments to pass to clang. See cautodoc_clang.')
+    parser.add_argument('--ingroup', metavar='PARAM',
+                        help='Only display the symbols matching ingroup. See cautodoc_ingroup.')
     parser.add_argument('--verbose', dest='verbose', action='store_true',
                         help='Verbose output.')
     args = parser.parse_args()
 
     comments = parse_to_string(args.file, args.verbose,
-                               compat=args.compat, clang=args.clang)
+                               compat=args.compat, clang=args.clang, ingroup=args.ingroup)
     sys.stdout.write(comments)
 
 main()
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -43,7 +43,7 @@ from clang.cindex import TokenKind, TokenGroup
 
 from hawkmoth.util import docstr, doccompat
 
-def comment_extract(tu):
+def comment_extract(tu, filterg):
 
     # FIXME: How to handle top level comments above a cursor that it does *not*
     # describe? Parsing @file or @doc at this stage would not be a clean design.
@@ -79,7 +79,11 @@ def comment_extract(tu):
 
         # Note: current_comment may be None
         if current_comment != None and docstr.is_doc(current_comment.spelling):
-            comments[cursor.hash] = current_comment
+            if filterg is not None:
+                if docstr.has_group(current_comment.spelling, filterg):
+                    comments[cursor.hash] = current_comment
+            else:
+                comments[cursor.hash] = current_comment
         current_comment = None
 
     # comment at the end of file
@@ -249,13 +253,14 @@ def parse(filename, **options):
         if len(args) == 0:
             args = None
 
+    ingroup = options.get('ingroup')
     index = Index.create()
 
     tu = index.parse(filename, args=args, options=
                      TranslationUnit.PARSE_DETAILED_PROCESSING_RECORD |
                      TranslationUnit.PARSE_SKIP_FUNCTION_BODIES)
 
-    top_level_comments, comments = comment_extract(tu)
+    top_level_comments, comments = comment_extract(tu, ingroup)
 
     result = []
     compat = lambda x: doccompat.convert(x, options.get('compat'))
--- a/hawkmoth/util/docstr.py
+++ b/hawkmoth/util/docstr.py
@@ -38,6 +38,15 @@ _doc_fmt = {
     Type.FUNC:       (1, '\n.. c:function:: {ttype} {name}({args})\n\n{text}\n')
 }
 
+def has_group(comment, filterg):
+    # we parse the entire comment and use either '@' or '\'
+    v = re.compile(r".*\n.*[@|\\]ingroup\s(\w+).*", re.MULTILINE | re.DOTALL)
+    m = v.match(comment)
+    # groups() returns the tuple while group(1) returns the actual value matched
+    if m and m.group(1) == filterg:
+        return True
+    return False
+
 def _strip(comment):
     """Strip comment from comment markers."""
     comment = re.sub(r'^/\*\*[ \t]?', '', comment)


jnikula:
Thanks for the contribution! I'm afraid this one goes directly against one of the (alas, not well enough documented) core design principles: as little parsing of the comments as possible.

Yes, we do have minimal and purely mechanical compat conversion code for Javadoc/Doxygen style `@tags`, but I'm actually hoping even that could be deprecated and moved into user-provided plugins.

Something like `@ingroup` is definitely out of the question. Everything about Hawkmoth should aim to align with Sphinx based solutions, and compatibility with Doxygen is of much lesser concern.

Now, with that out of the way, we should try to think of ways to achieve what you want using Sphinx!

BrunoMSantos:
I fully agree with Jani on this one. As a matter of fact, I don't think this should go in in any form. One of the main advantages of Sphinx over Doxygen is the much more powerful documentation composition, and this would compromise it. I also think composition belongs outside of the code anyway, even if it tends to being a little bit more verbose compared to Doxygen's approach.

My suggestion is to add directives that take a file and a (list of) function or variable name(s) and document only that to support the same level of granularity Sphinx offers for Python or that of Doxygen's `ingroup` but with the true versatility of Sphinx composition behind it as the user will have to provide the glue documentation that stitches it together.

I don't think it's a feature I'm missing personally, but I can see the advantage and it would bring hawkmoth's interfaces closer to Python's own autodoc.

mv0:
Hi,

Thanks for the replies. `ingroup` is just a keyword here. Could've been `tag` or something like `memberof` (I thought that using an analogy might be easier). I don't think it is worth mentioning doxygen at all, we all believe doxygen is too much to bear on. On the same idea, we do not seek to maintain compatibility with doxygen but the we'd like the ability to group certain symbols. Even `kerneldoc` has this ability to display whatever symbols you want. 

Personally do not see how the ability to specify which symbols you want display hinders hawkmoth design principles and this small feature far outweighs the "complexity" it adds by making it aware of a certain keyword in the comments. 

@jnikula I'm guess you're hinting to look into Sphinx C Domain to add this functionality? I see more reasons from their part to reject this: 1. There are already doxygen bridges (breathe) 2. Why add just this functionality and not all the others present in doxygen? 3. Why not use doxygen directly with some custom reST parser.

jnikula:
I emphasize that the main argument here is that I want to keep the parsing of the comments to a bare minimum: See if the comment starts with `/**` and strip the comment markers. Everything else gets passed to Sphinx verbatim.

This is one of the key lessons from Linux `kernel-doc`. All the pre-processing of the comments leads to hard to solve problems. That was a necessary evil in `kernel-doc` due to the massive amounts of existing comments, but Hawkmoth was born from those frustrations with the thought, "How should this be done if I were free from legacy baggage?"

So any solution adding grouping or selective inclusion of documentation must do it without parsing the comments. That's pretty much non-negotiable.

---

One obvious idea is to add some sort of symbol filters in the `c:autodoc` directive. Alas, I think we got this part wrong in the Linux kernel `kernel-doc` directive extension too. It's not coherent, perhaps largely because it was not coherent in `kernel-doc` the Perl script either. Worse, it has lead to somewhat poor form in documentation: picking some symbols here, some others there, and leading to multiple parser invocations per source file due to the multiple directives with different filters. I'd like to avoid these mistakes in Hawkmoth.

Finally, I have yet to look at Sphinx for ideas how to implement what you want in a "sphinxy" way.

I guess that's a long-winded way of saying, I'm not dismissing the idea, I'm dismissing the implementation.

BrunoMSantos:
I may have explained myself poorly. My own suggestion was about providing directives like `automodule` (whole file), `autofunction` (single function), etc. instead of our `autodoc`. À la https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html.

Doing so would support grouping, but, crucially, the grouping is kept _outside_ of the source code. The kernel supports this as well, but with a single directive and extra parameters to filter the output (e.g.: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/crypto/api-aead.rst?h=v5.1-rc6). Problem I see with the kernel's approach is that it doesn't fit nicely next to upstream's `autodoc` API wise.

Your patches are similar to the kernel's approach in that you add parameters to the directive to filter the output, but worse in my view because instead of filtering on symbols it filters on tags that must be parsed. Doing that you effectively spread documentation grouping and structuring (what I mean by composition) across `.rst` and `.c/h` files.

In summary, Sphinx/reST already allows us to create sections for structuring the document. We only need more discriminative directives or more parameters to the one we have to tap into that (I vote for the former).

jnikula:
> In summary, Sphinx/reST already allows us to create sections for structuring the document. We only need more discriminative directives or more parameters to the one we have to tap into that (I vote for the former).

I mostly agree. It's probably least suprising to emulate autodoc here. The one issue I see is the Clang invocation per directive, while I believe autodoc for Python imports the module just once.

BrunoMSantos:
Hm, that is a good point. Of course it's easy for Python due to its dynamic and introspection capabilities... We should be able to emulate that behaviour by caching the results.

We could also start by taking the performance hit initially and add caching as a later improvement. It will only affect those highly separating stuff anyway.

mv0:
Thanks for taking the time to comment extensively :-) on this matter. 

> I may have explained myself poorly. My own suggestion was about providing directives like automodule (whole file), autofunction (single function), etc. instead of our autodoc. À la https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html.

@BrunoMSantos Isn't this python related? Assuming one "modifies" this Sphinx extension how would that be helpful? I'm quite a novice into Sphinx so I'm probably missing some key points. I gather that Hawkmoth is a for the C language, what sphinx.ext.autodoc is for python language. I also gather that the former already knows to filter the symbols one would want to generate reST part (autofunction seem to indicate that). To that matter, based on both of your comments the only way this would be added is by modifying Sphinx itself, but given that Sphinx was/is design for documenting Python how would one go  about it?

BrunoMSantos:
Yes, Python only, I was pointing at the interface only. Also, Sphinx
already supports several languages through different domains
(https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html),
including C, so there's no reason to touch Sphinx code itself.


Maybe this explains better what I meant (note that referring to symbols
directly removes the need for tags):

```rst
Whole file (what we currently have)
===================================

.. c:autodoc:: path/to/main.c

Group one
=========

.. c:autodata:: path/to/main.c some_variable

Group two
=========

.. c:autofunction:: path/to/main.c main
.. c:autofunction:: path/to/main.c another_function

About the 2 parameters
======================

In Python we can do this, virtue of its dynamic nature and introspection
capabilities:

.. py:autofunction:: fun
.. py:autofunction:: module.fun

Closest I think we could come to that is:

.. c:autofunction:: path/to/main.c/another_function

But I don't like that...
```

------------------------------------------

Before we keep discussing implementation though, I think we need to
consider something else. @jnikula raised a point about this leading to
poor documentation form in the kernel.

Initially I was thinking 'with great power comes great responsibility'
and not particularly concerned about its evil uses, but the more I think
about it, the more I realize I have no real scenario to justify doing
this in the first place. Maybe you guys can think of one?

These are the scenarios I can picture together with my opinionated
reasoning about them:

1. Separating 'public' and 'private' stuff or different levels of API
   cruft. However, a better way is to keep some code hygiene: private
   stuff should not be in a public header.

   Also, the tendency to document every single function and variable is
   wrong and facilitating that is not in our interest. Some stuff just
   shouldn't be documented at all (with docstrings anyway).

   *-1*

2. Separating documentation per configuration. This is done with the
   pre processor or by including different files anyway so this is not
   covered by this new grouping stuff and is already supported with our
   single directive.

   *-1*

3. Separating deprecated symbols, etc. This should be done in the code
   in native Sphinx syntax instead:
   https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

   *-1*

4. Split a module in several subsections. This is possible already except
   it doesn't allow to break this down or change its order. The source
   code rules composition in this case, but I think that's still
   tasteful as it does so as a whole and in a very straightforward way
   with no extra syntax elements:


   ```c
   /**
    * Section 1
    * =========
    *
    * Note you can write reST to your heart's content here.
    */
   
   /**
    * And here...
    */
   void fun_1(void);
   
   /**
    * Section 2
    * =========
    *
    * Meta question to entertain the wandering mind: would Sphinx
    * recurse here?
    *
    * .. c:autodoc:: this_file.c
    */
   
   /**
    * ...
    */
   void fun_2(void);
   ```

   Note this is not possible for Python as Python enforces a single
   documentation block at the top of the file. You can add more
   documentation blocks but they will not be extracted into the module's
   `__doc__`. Don't think this should matter though.

   *-1*

5. Grouping symbols from different files onto the same group. I don't
   think this is good form unless we can include both files completely
   under a common top level, which is already supported:

   ```rst
   Group 1
   =======
   
   .. c:autodoc:: file1.c
   .. c:autodoc:: file2.c

   Maybe a sub category
   --------------------

   .. c:autodoc:: file3.c
   ```
   
   *-1*

6. Some combination of 4 and 5 where this would be expected:

   ```rst
   Group 1
   =======
   
   .. c:autofunction:: file1.c fun1
   .. c:autofunction:: file2.c fun1

   Group 2
   =======
   
   .. c:autofunction:: file1.c fun2
   .. c:autofunction:: file2.c fun2
   ```

   At this point I'd start wondering what kind of spaghetti monster this
   code is if parts of separate files belong together better than they
   belong with other parts of the same file...

   *-1*

Finally, I don't think I would ever use the fine grained interfaces myself,
but I'm willing to support the idea if a valid case is suggested. In any
case, @jnikula's experience from the kernel may be the most relevant we have
so far, so I'd be inclined to follow his lead.

ppaalanen:
I kind of see @BrunoMSantos' point. It is equally easy to forget to tag a symbol in its doc comment as it is to forget to add the symbol in a group list somewhere else.

Think about how C++ groups (public) symbols under a `class`. Is there a way to find all symbols that do not belong to any "class" in C? Class in C is a special kind of grouping, just like `class` is in C++.

Is there a way to pick up arbitrary symbol grouping by header files while the symbol doc comments are in `.c` files?

BrunoMSantos:
> Is there a way to pick up arbitrary symbol grouping by header files while the symbol doc comments are in .c files?

No, you need to place public documentation in the header file and do `.. c:autodoc:: public.h`, which I believe is the tastier option anyway. Category 1 :)

jnikula:
> > Is there a way to pick up arbitrary symbol grouping by header files while the symbol doc comments are in .c files?
> 
> No, you need to place public documentation in the header file and do `.. c:autodoc:: public.h`, which I believe is the tastier option anyway. Category 1 :)

I think a fair argument is that documentation comments in the .c files are more likely to get updated than in the .h files. I think both approaches should be fine.

However using the contents of one file to adjust the parsing of another is something I'd steer clear from.

BrunoMSantos:
> I think a fair argument is that documentation comments in the .c files are more likely to get updated than in the .h files. I think both approaches should be fine.

Well yes, but as a user who only has the headers installed in his search path, I'd want the documentation to be in the header where I can find it and prototypes can also fall out of sync. But fair point indeed. I'm willing to accept this as a good argument against point 1.

Where does your comment about such features leading to poor documentation form leaves us though?

jnikula:
Some thoughts on grouping:

- Arrange the source in a way that reflects the desired grouping (what we have now).

- Add ability to **sort** the symbols by source location (i.e. what we have now), symbol name, type, attributes, etc. Would only sort within a file. Multiple groups spread across multiple files does not work. Includes all documented symbols.

- Add ability to **filter** the symbols by name, type, attributes, etc. Could use regex or globbing. Need to use multiple directives to group stuff. Currently leads to Clang invocation per directive. Clutters the Sphinx side of the documentation, may miss documented symbols, may document some symbols more than once. Allows grouping across files.

- Figure out how to leverage existing Sphinx indexing mechanisms, and/or add another extension for generating indexes or table-of-contents based on some metainformation added to the documentation. Admittedly this is a bit hazy to me, would need to study Sphinx more.

ppaalanen:
> No, you need to place public documentation in the header file and do `.. c:autodoc:: public.h`, which I believe is the tastier option anyway. Category 1 :)

When documentation is separated from the implementation, the likelyhood of it being out-of-date is significantly higher, if it even exists in the first place. That's why I don't like it.

I also think that properly generated docs are to be installed. End users reading the source code of docs (e.g. doc comments in headers) is always going to be a poorer experience than reading the actual docs the way they are meant to be.

> However using the contents of one file to adjust the parsing of another is something I'd steer clear from.

Does it have to be about changing parsing? Can't parsing be done once and for all, and then the resulting database be used and filtered in various ways? Or is that just not how Sphinx works?

jnikula:
> > However using the contents of one file to adjust the parsing of another is something I'd steer clear from.
> 
> Does it have to be about changing parsing? Can't parsing be done once and for all, and then the resulting database be used and filtered in various ways? Or is that just not how Sphinx works?

The extension registers a new directive to Sphinx (`.. c:autodoc::`). When Sphinx encounters the directive in .rst source, our hook gets called. We run Clang on the source file given as parameter to the directive, and hand back the rst generated from comments to Sphinx. This happens on every directive, regardless of whether the file is the same.

Currently we don't do any caching of the parse results. We parse and throw away the tokens and the AST. This is pretty straightforward.

Adding a "database" or caching the results is up to us. I can imagine this could be done at multiple levels. For example caching just the parsed contents per source file, or caching the tokens and the AST. Especially the latter sounds problematic in big projects. All in all this is not something that should be anywhere near the top of the TODO list of the project. Simplicity is key.

BrunoMSantos:
Ok guys, so if we're doing this, I'd personally go for the kind of thing I illustrated before at the top of https://github.com/jnikula/hawkmoth/pull/9#issuecomment-485980352: multiple directives to pick different types of symbols at will and renaming `autodoc` to `automodule` while at it.

I'd also avoid regexes or globbing... But if there's a demand for it I won't oppose it. Note Python's `autodoc` supports neither as far as I know.

Agree with postponing the caching until a later date. We have bigger fish to fry.

ppaalanen:
I see, so it's just not how Sphinx works.

I have to chat with Marius about what we would actually need, because I'm getting the feeling we might not need everything mentioned here.

jnikula:
> I have to chat with Marius about what we would actually need, because I'm getting the feeling we might not need everything mentioned here.

Please document your needs in github issues, and let's then try to figure out what would be the cleanest way to address those needs. Thanks.

ppaalanen:
Here is an idea for the alternative approach: https://gitlab.freedesktop.org/wayland/weston/issues/216#note_151616 which I think is very close to what @BrunoMSantos talked about.

I also just realized we'd need two different flavours for including a function's doc: the full doc in some place, and then something more suitable for an index, like the doc of a `struct` listing all related functions in an index, followed by the full docs. Maybe something like this: https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html

Would this be achievable somehow?

I probably would not rule out the option of making one header file per "class" to achieve this, but I'd like to hear about alternatives first. So far our public headers in libweston are far bigger than one class, more like a file per module.

jnikula:
It's a bit more manual work, but already now you can achieve a similar effect by documenting all your functions, adding one `.. c:autodoc::` directive per source file, which generates the documentation as usual, and then reference the relevant functions either in a struct or generic documentation comment, using regular Sphinx cross-referencing syntax. The cross-reference targets are automatically generated for everything Hawkmoth produces.

For example:

```
/**
 * A surface with client or internal content
 *
 * Blah blah...
 *
 * Methods:
 *
 *   - :ref:`weston_surface_create`
 *   - :ref:`weston_surface_get_output`
 *   - :ref:`weston_surface_unmap`
 *   - :ref:`weston_surface_destroy`
 */
struct weston_surface;
```

The documentation for each function would be inserted from the documentation comments at the `.. c:autodoc::` directive, in the order they appear in the file, the file is only processed once. The manual work is adding the references to the relevant methods in the struct documentation.

It's not perfect, but on the upside it works *now*.

For more info on referencing, see:

http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-c-constructs
http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#xref-syntax

jnikula:
The request has gone stale. I think something like this might be a nice feature in the long run.

PR #10, opened 2019-04-24
Add way of raising error messages from the parser

BrunoMSantos:
This follows the current pattern used for passing the actual parsed
results except in a separate tuple. Currently it adds only diagnostics
from Clang, but the same concept can be extended to pass other warnings
or errors (e.g.) when we find weird documentation so we prod the user in
the right direction.

Original concept: Marius Vlad

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -46,7 +46,10 @@ class CAutoDocDirective(Directive):
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
 
-        comments = parse(filename, compat=compat, clang=clang)
+        comments, errors = parse(filename, compat=compat, clang=clang)
+
+        for (ffile, line, msg) in errors:
+            env.app.warn(_err_fmt.format(msg, location=(ffile, line)))
 
         for (comment, meta) in comments:
             lineoffset = meta['line'] - 1
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -239,6 +239,15 @@ def _recursive_parse(comments, cursor, nest, compat):
 
     return [(doc, meta)]
 
+def _clang_diagnostics(tu):
+    errors = []
+    for d in tu.diagnostics:
+        # Filter for warnings and errors
+        if d.severity >= 2:
+            errors.extend([(d.location.file.name, d.location.line, d.spelling)])
+
+    return errors
+
 # return a list of (comment, metadata) tuples
 # options - dictionary with directive options
 def parse(filename, **options):
@@ -255,6 +264,8 @@ def parse(filename, **options):
                      TranslationUnit.PARSE_DETAILED_PROCESSING_RECORD |
                      TranslationUnit.PARSE_SKIP_FUNCTION_BODIES)
 
+    errors = _clang_diagnostics(tu)
+
     top_level_comments, comments = comment_extract(tu)
 
     result = []
@@ -270,7 +281,7 @@ def parse(filename, **options):
     # Sort all elements by order of appearance.
     result.sort(key=lambda r: r[1]['line'])
 
-    return result
+    return result, errors
 
 def parse_to_string(filename, verbose, **options):
     s = ''


hawkmoth/__init__.py
```diff
@@ -46,7 +46,10 @@ def __parse(self, viewlist, filename):
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
 
-        comments = parse(filename, compat=compat, clang=clang)
+        comments, errors = parse(filename, compat=compat, clang=clang)
+
+        for (ffile, line, msg) in errors:
+            env.app.warn(_err_fmt.format(msg, location=(ffile, line)))
```
BrunoMSantos: This doesn't work right now (not sure why actually, but didn't dwell on it either, this is an RFC).

@jnikula I actually don't know where this comes from, but I saw it a few lines below in `run` and thought it might be the correct way to pass warnings/ errors to Sphinx.

hawkmoth/parser.py
```diff
@@ -239,6 +239,15 @@ def _recursive_parse(comments, cursor, nest, compat):
 
     return [(doc, meta)]
 
+def _clang_diagnostics(tu):
+    errors = []
+    for d in tu.diagnostics:
+        # Filter for warnings and errors
+        if d.severity >= 2:
```
BrunoMSantos: I don't know if we should be passing so many things back to the user. This will give warnings like 'symbol not defined'. Well... we shouldn't replace the build system. If we call the directive with a single source file, it's only natural a bunch of definitions will be missing.

Haven't found documentation on what each level means though.

mv0: Yep. You're right. Documentation is 0 :)

Here's what sources says:
```
    Ignored = 0
    Note    = 1
    Warning = 2
    Error   = 3
    Fatal   = 4
```

BrunoMSantos: I disagree. I'd go for error/fatal only, whichever relates to Clang bailing out on us altogether. E.g. 'missing definitions' is to be expected under normal operation which renders the output uselessly noisy at level 2.

Sphinx should be printing errors found in documentation (function parameters missing in the docstring or not matching, or even more fundamental bad uses). I think it's good to print something about Clang failing completely, but the user should have actually built the code to find source errors.

I just don't know what's the difference between error and fatal exactly.

mv0: Fatal is missing files. When doing the AST it recursively reads header files. If it can't find it it will go with severity 4 (Fatal).

mv0: BTW, when this happens the AST parser will silently stop processing symbols. The end result is that we end up comments in the documentation but no cursor (symbol declaration).

BrunoMSantos: Hmm, maybe I'm wrong then. Plenty of our test cases are incomplete sources and yet they produce good documents though. I assumed missing headers would have a similar result, but haven't actually tried it.

For instance, I wouldn't want to do this in the documentation if possible:
```rst
.. c:autodoc:: source.c
   :clang: -Ipath/to/headers
```

mv0: Yep, that would be need for some of the tests. See my initial PR where I've modified a bit the test scripts and how it prints.

mv0: I've tested a bit the sphinx methods of logging/printing:

- info prints without any prefix 
- debug doesn't print all all (I assume that this has to be enabled)
- critical/fatal/warning methods all add a prefix/prepend with `WARNING: ` string before. I'd expected these to print different but it doesn't appear to be the case?

BrunoMSantos: Crap... If there's no way around it then I think we must add all those warnings and errors of course. The user should know why the documentation build failed.

Maybe there's something in how we invoke Clang that can make it ignore missing files. If we cut the parsing at different levels it may affect how much information we can extract from the code though, so I'm not very optimistic.

In the meantime I'm closing this PR, I think it served its purpose anyway and you've kept updating your own. Let's consolidate the discussion over there.

BrunoMSantos:
Following up in original PR.

jnikula:
Implemented by commits:

bd152fc4ad6a ("hawkmoth: add support for propagating diagnostics")
215963c57df9 ("test: add diagnostic tests")

PR #15, opened 2019-06-11
Standardize newlines to be able to run test on windows machines

speedyleion:
This was only proposed in the mailing list and I know there may still be discussion on whether this is the best/correct way to handle newlines across OS's.

P.S. The other test failures I had mentioned in the mailing list was git/windows/symlinks and those have been addressed at the user level.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -277,8 +277,15 @@ def parse(filename, **options):
 
     index = Index.create()
 
-    tu = index.parse(filename, args=args, options=
-                     TranslationUnit.PARSE_DETAILED_PROCESSING_RECORD |
+    # In order to let python standardize line endings across operating systems
+    # by reading in the file contents and sending that to libclang as an
+    # `unsaved_file`.
+    with open(filename, 'r') as f:
+        contents = f.read()
+        unsaved_files = [(filename, contents)]
+
+    tu = index.parse(filename, args=args, unsaved_files=unsaved_files,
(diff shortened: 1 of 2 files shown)


jnikula:
What's the failure mode that this fixes? I'd very much like to let clang read the source files itself.

speedyleion:
I've pushed up a second attempt, cleaning the newlines in [_strip()](https://github.com/jnikula/hawkmoth/blob/3822405c5f0f71fedf1feb7102a4f7cc800927a5/hawkmoth/util/docstr.py#L41).

Note: the first commit was a force push to reset the author per the comment in https://github.com/jnikula/hawkmoth/pull/16

This is meant to fix the following kinds of test failures:
The expected test files are opened in universal newline mode, and the `_strip()` regex doesn't handle the `\r\n` line endings.
```
FAIL: test_struct (__main__.ParserTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\git\hawkmoth\test\testenv.py", line 80, in test
    self.assertEqual(expect_docs, output_docs)
AssertionError: '\n..[56 chars]ruct\n\n   Woohoo.\n\n\n   .. c:member:: int j[87 chars]\n\n' != 
                '\n..[56 chars]ruct\r\n   \r\n   Woohoo.\n\n\n   .. c:member:[94 chars]\n\n'

  .. c:type:: struct sample_struct

-    This is a sample struct
+    This is a sample struct
?                           +
-
+
     Woohoo.
```

PR #16, opened 2019-06-11
Re-use the looked up cursor of a token.

speedyleion:
A token's cursor is a property which is looked up every time it's
accessed.  Since the token's cursor doesn't change it is kept in a local
variable and re-used for a small performance benefit.

I understand speed may not be a concern at this point, but I know from experience using `token.cursor` a lot ends up being a bottleneck and I don't think the code becomes obfuscated with this change.
* `test_hawkmoth.py` went from ~2.1 sec to ~1.7 seconds on my machine.
* `test_cautodocpy` went from ~6.6 sec to ~5.9 sec on my machine.

If you look at [cindex.Token.cursor](https://github.com/llvm-mirror/clang/blob/master/bindings/python/clang/cindex.py#L3308) You can see that it's a property which is looked up each time.  I haven't looked into the clang internals, but my guess is that it walks the  AST looking for the extent that the token lies in.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -75,18 +75,22 @@ def comment_extract(tu):
             current_comment = token
             continue
 
+        # Store off the token's cursor for a slight performance improvement
+        # instead of accessing the `cursor` property multiple times.
+        token_cursor = token.cursor
+
         # cursors that are 1) never documented themselves, and 2) allowed
         # between comment and the actual cursor being documented
-        if (token.cursor.kind == CursorKind.INVALID_FILE or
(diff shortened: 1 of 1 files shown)


jnikula:
Sorry for taking so long to get around to this. I'm swamped.

The change looks good to me. But could I persuade you to set your real name and mail address to the commit author?

speedyleion:
Updated author info.
Note: this is a force push in order to update the commit message.

Thanks!

jnikula:
Merged, sorry for taking approximately forever to get around to this.

PR #30, opened 2021-01-28
Function pointer fix v2

jnikula:
Improved version to take into account pointers to function pointers with any level of indirection, as well as arrays of the same.

And now as a pull request so it's possible to actually comment. :)

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -166,8 +166,11 @@ def _recursive_parse(comments, cursor, nest, compat):
         return _result(comment, cursor=cursor, fmt=fmt,
                        nest=nest, name=name, args=args, compat=compat)
 
-    elif cursor.kind == CursorKind.VAR_DECL:
-        fmt = docstr.Type.VAR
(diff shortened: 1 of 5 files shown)


hawkmoth/parser.py
```diff
@@ -172,12 +173,21 @@ def _recursive_parse(comments, cursor, nest, compat):
...
+        # the type.
         dims = ttype.rsplit(' ', 1)[-1]
         if dims.startswith('[') and dims.endswith(']'):
             ttype = ttype.rsplit(' ', 1)[0]
             name = name + dims
 
+        # If this is a function pointer, or an array of function pointers, the
+        # name should be within the parenthesis as in (*name) or (*name[N]).
+        fptr_type = re.sub(r'\((\*+)(\[[^]]*\])?\)', r'(\1{}\2)'.format(name),
+                           ttype, count=1)
```
BrunoMSantos, line 186: Nice! Thanks for addressing this.

BrunoMSantos (approved):
I think it's good to go :)

PR #31, opened 2021-01-28
util: add helper for setting up libclang on readthedocs

jnikula:
There's a bit of a chicken-and-egg problem for testing this, as RTD would install hawkmoth from pypi, which does not have the change... Also the required version in the documentation is set as >=0.7 which is unreleased. However I tested this with a dummy project with the function copied over:

https://github.com/jnikula/curly-octo-tribble
https://curly-octo-tribble.readthedocs.io/en/latest/

(curly-octo-tribble was a random name github suggested :)


jnikula:
Decided to just go ahead and merge this to get the release done, as there's really no regression potential here.

PR #33, opened 2021-03-06
Transformations RFC

jnikula:
Not intended for merging as-is, but here's some ideas of what to do with compat and transformations, hoping to get some feedback.

Basically, start deprecating "compat" and moving towards user-defined "transformations". The transformations can reuse the existing doccompat functions, extend them, support fully user-written transformations, etc.

There are probably enough users to warrant some level of backwards compatibility while deprecating compat. But long term I'd like Hawkmoth focus to be on basically passing through the comments from clang to sphinx with minimal C comment mark and indentation removal, with no extra conversion that can go wrong.

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -67,8 +67,20 @@ class CAutoDocDirective(Directive):
 
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
+        transformations = env.config.cautodoc_transformations
 
-        transform = lambda x: doccompat.convert(x, mode=compat)
+        if transformations is not None:
+            # compat == 'none' works by coincidence if there is no 'none'
+            fn = transformations.get(compat, None)
+            if fn is not None:
+                transform = lambda x: fn(x, mode=compat)
+            else:
+                transform = None
+        elif compat is not None:
+            # FIXME: deprecated fallback
+            transform = lambda x: doccompat.convert(x, mode=compat)
+        else:
+            transform = None
 
         comments, errors = parse(filename, transform=transform, clang=clang)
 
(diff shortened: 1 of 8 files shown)


BrunoMSantos:
Sorry for taking so long to look into this. I think this is a brilliant concept and the napoleon use case is one I actually care about as I've recently started using it in some Python code.

I think this should indeed go in and any current options (napoleon as well I guess) left as examples, but not included in the whole thing by default. That said I may have a few comments on the actual code, but I want to try a couple of things out before I post them. Just wanted to let you know I haven't forgotten about it given the delay ;)

hawkmoth/__init__.py
```diff
@@ -67,8 +67,20 @@ def __parse(self, viewlist, filename):
...
         compat = self.options.get('compat', env.config.cautodoc_compat)
         clang = self.options.get('clang', env.config.cautodoc_clang)
+        transformations = env.config.cautodoc_transformations
 
-        transform = lambda x: doccompat.convert(x, mode=compat)
+        if transformations is not None:
+            # compat == 'none' works by coincidence if there is no 'none'
+            fn = transformations.get(compat, None)
+            if fn is not None:
+                transform = lambda x: fn(x, mode=compat)
```
BrunoMSantos, line 76: It's weird to derive a function from `compat` above only to then call a different entry point with `mode=compat`. It seems we can get rid oif the `convert()` function and do directly `transform = transformations.get(compat, None)`

BrunoMSantos (commented):
Again, sorry for taking so long to reply. I think this is a pretty solid move and a nice RFC. I don't even think there's much to change before we consider it for merging.

hawkmoth/__init__.py
BrunoMSantos: Actually I was partially off, I was confused by the passing of `compat` into the function. Maybe a remnant of using `convert()`?

Unless we want to pass additional options to the convert function from the user (probably a good and easy thing to do), there's still no need to capture this lambda though.

jnikula:
@BrunoMSantos Thanks for the comments, and sorry for not replying. I was trying to come up with a clean deprecation path from compat to transformations, as well as fix the testing.

PR #38, opened 2021-04-10
Parser fixes

jnikula:
Address a couple of recent parser bug reports.

### DIFF
--- a/test/function.c
+++ b/test/function.c
@@ -12,3 +12,15 @@ int fooing(void);
  * Empty parameters.
  */
 int fooation();
+
+/**
+ * Array parameters.
+ */
+void foorray(const double array[], int x[5]);
+
+/**
+ * Array parameters with multiple dimensions.
+ *
+ * Clang removes spaces.
+ */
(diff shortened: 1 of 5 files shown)


test/function.c
```diff
@@ -12,3 +12,8 @@ int fooing(void);
  * Empty parameters.
  */
 int fooation();
+
+/**
+ * Array parameters.
+ */
+void foorray(const double array[], int x[5]);
```
BrunoMSantos, line 19: Maybe we can add a couple of more cases for good measure: `[1][2]` and `[1] [2]` (note the space). These should be ok unless clang introduces / keeps spaces between the bracket elements.

BrunoMSantos (approved):
Nice clean-ups ;)

Had one minor comment about additional test cases, but I think it's pretty solid.

test/function.c
jnikula: Thanks, good idea, added and updated the branch.

jnikula:
And merged, many thanks for the review!

Issue #29, opened 2021-01-27
Include files not found


LudovicGuerra:
Clang doesn't find some include file (which is normal), and show an error like this:

> C:\Users\*\Desktop\Code\f1\ula\doc\drivers\control\RTC\drv_rtc_pcf85063_api.rst:24: WARNING: ../include\sys/type.h:21: 'inttypes.h' file not found

If I give clang the folder where inttypes.h is, I will have a cascade of error (redefinition, etc)
I searched through  clang parameters (https://releases.llvm.org/6.0.0/tools/clang/docs/ClangCommandLineReference.html) but I didn't find any to ignore specific includes.

Do you have any idea how I could handle this problem ?

jnikula:
Is it really supposed to be `sys/type.h` and not `sys/types.h`?

Have you tried `clang` on the file directly, without Sphinx or Hawkmoth?

LudovicGuerra:
Yes as sys/type.h is our own file.
No I didn't tried clang directly, as I expect it to have the same error. I used the parameters -I to give our include files. But when clang read the files, It search some include file (as inttypes.h) which are located somewhere else in the system, but if I add them I have too many error in cascade as no macro are defined for example. I'd like to simply says: "Please ignore inttypes.h include and don't try to parse it"

jnikula:
I'm at a point where I'm just guessing...

I'd think clang should be able to find its own standard includes. That smells like a clang installation problem to me. On the other hand, when you're adding include paths with `-I`, are you pointing them at clang standard includes or something else? It's quite possible e.g. visual studio standard includes won't work.

wkschwartz:
I was having similar problems on macOS:

```
<omitted>.rst:13: WARNING: <omitted>.h:6: 'stddef.h' file not found
```

Appending `-I/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include` to `cautodoc_clang` in `conf.py` got rid the errors except for missing `stdarg.h`, which doesn't appear to be in that directory.

```console
$ python --version
Python 3.7.10
$ pip list | grep clang
libclang                      11.1.0
$ clang --version
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

jnikula:
Turns out `libclang` does not infer the system include directories the way `clang` on the command line does.

[...]

PR #39, opened 2021-04-11
RFC: Discover system include path

jnikula:
Fix for #29.

I've got mixed feelings about this.

On the one hand, it's an obvious, fairly isolated and straightforward thing to do. It should make Hawkmoth easier to use for real users out there. And I've written the code, it's there.

On the other hand, I clearly have a configuration where I can't reproduce the issues (I don't know why; maybe recent libclang already does this?), I have a hard time thinking of reasonable and portable test cases for this, and it also has a feeling of bloat beyond the KISS principle.

WRT design, perhaps this should be handled at one abstraction level higher than the parser. Since the parser is all about clang and its python bindings, and this uses the compiler front-end directly.

### DIFF
--- a/doc/extension.rst
+++ b/doc/extension.rst
@@ -67,6 +67,20 @@ The extension has a few configuration options that can be set in ``conf.py``:
    source, typically to define macros for conditional compilation, for example
    ``-DHAWKMOTH``. No arguments are passed by default.
 
+.. py:data:: cautodoc_cc_use_system_include_path
+   :type: bool
+
+   Try to use the compiler specified by :data:`cautodoc_cc_path` to figure out
+   the system include paths, and pass them as ``-I`` options for parsing the
+   source code comments. False by default.
+
+.. py:data:: cautodoc_cc_path
+   :type: str
+
+   Compiler to use for figuring out the system include paths if
+   :data:`cautodoc_cc_use_system_include_path` is enabled. This is not used for
+   parsing the source code comments. Defaults to ``clang``.
+
 Directive
 ---------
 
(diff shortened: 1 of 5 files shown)


BrunoMSantos:
I had only a quick look. I confess I have the same concerns. I do a lot of embedded / bare metal stuff and this would not be that generic there. And the support for finding headers with other compilers is even worse as clang may choke on headers that require some (e.g.) `__GNUC__` define to declare that super important type we depended on or whatever. Not to mention compilers that use non gcc compliant switches of course.

So, to my mind it makes more sense to assume nothing and rely on the user to do the work for consistent results and figure out any workaround if mixing compilers and so on. But, not being the target audience, I feel disqualified to vote it down entirely. 

I also think issues like #29 will keep on coming regardless of whether we do this or not. But they will be less frequent if we do it. Maybe that's all we need?

jnikula:
Maybe I'm just adding this way too deep in the guts of Hawkmoth? What if, instead, we added an independent helper in `hawkmoth.util` that the users could call in their `conf.py` to add to `cautodoc_clang`? Similar to `hawkmoth.uti.readthedocs`.

Looking at this, seems a bit silly that the directive `clang` option *overrides* the `cautodoc_clang` config option. It should probably append to it instead. And it should probably be okay to add `cautodoc_clang` directly as a list instead of string.

This way, the core of Hawkmoth would stay pretty clean, and without additional configuration too.

BrunoMSantos:
That sounds good indeed, although I confess I didn't read through these patches with an eye on bigger integration issues, I was only looking at the functionality itself.

PR #45, opened 2021-04-29
Packaging update

jnikula:
Move towards the recommendations in https://packaging.python.org/tutorials/packaging-projects/


BrunoMSantos (approved):
Nice, didn't know about these recommendations. I may have to update the Arch AUR packages for this, but the changes look good!

PR #46, opened 2021-04-29
Transformations part 1

jnikula:
I got a bit stuck with adding tests for the transformations (see #33), but here's the preliminary work to clean stuff up, so I don't have to carry it locally.

### DIFF
--- a/hawkmoth/util/doccompat.py
+++ b/hawkmoth/util/doccompat.py
@@ -17,46 +17,71 @@ import re
 # facilitate any kind of migration to Hawkmoth. The compat code could be turned
 # into a fairly simple plugin architecture, with some basic compat builtins, and
 # the users could still extend the compat features to fit their specific needs.
+#
+# FIXME: try to preserve whitespace better
+#
+
+def javadoc(comment, **options):
+    """Basic javadoc conversion to reStructuredText"""
+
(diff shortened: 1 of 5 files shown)


hawkmoth/util/doccompat.py
```diff
@@ -17,11 +17,13 @@
 # facilitate any kind of migration to Hawkmoth. The compat code could be turned
 # into a fairly simple plugin architecture, with some basic compat builtins, and
 # the users could still extend the compat features to fit their specific needs.
```
BrunoMSantos, line 19: Hmm, I was comparing with the RFC notes and I noticed this `FIXME` is still here. Is it not obsolete with this series?

BrunoMSantos (approved):
Looks good to me :)

hawkmoth/util/doccompat.py
jnikula: Not yet. I have the implementation pretty much ready locally, but I had some issues testing it.

jnikula:
Thanks again for your review!

hawkmoth/util/doccompat.py
BrunoMSantos: Alright. looking forward to it then ;)

PR #48, opened 2021-04-30
System include path v3

jnikula:
Superseeds: #44 

Some refactoring over the previous series. Main difference is converting the user facing `cautodoc_clang` configuration to a list. (It may still be passed as string to `parse()` due to tests, but that cleanup is for another time.)


BrunoMSantos (approved):
Again, nice patches. My one issue with this remains the comma separation of arguments, which I think should be white space separation instead. But as discussed before, that's a minor thing, especially now that we can use a list instead, and I think you already weighted that decision.

Good to go ;)

jnikula:
Thanks for the review!

PR #49, opened 2021-05-04
Transformations v2

jnikula:
Transformations, finally. :)

I'd like to get this and #48 merged and then do a release.

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -38,6 +38,7 @@ class CAutoDocDirective(Directive):
     final_argument_whitespace = True
 
     option_spec = {
+        'transform': directives.unchanged_required,
         'compat': directives.unchanged_required,
         'clang': directives.unchanged_required,
     }
@@ -62,14 +63,53 @@ class CAutoDocDirective(Directive):
                 self.logger.log(self._log_lvl[severity], toprint,
                                 location=(env.docname, self.lineno))
 
-    def __get_transform(self):
+    def __get_compat_transform(self):
         env = self.state.document.settings.env
 
         compat = self.options.get('compat', env.config.cautodoc_compat)
         if compat is None:
             return None
 
+        fmt = 'cautodoc_compat and compat options are deprecated, please use cautodoc_transformations and transform options instead.'
+        self.logger.warning(fmt, location=(env.docname, self.lineno))
+
         return lambda comment: doccompat.convert(comment, transform=compat)
+
+    def __get_transform(self):
+        env = self.state.document.settings.env
+
+        # Special case values for transform.
+        OPT_NONE = 'none'
+        OPT_DEFAULT = 'default'
+
+        # Handle deprecated compat. To be removed.
+        transform = self.__get_compat_transform()
+        if transform is not None:
+            return transform
+
+        transformations = env.config.cautodoc_transformations
+        tropt = self.options.get('transform', OPT_DEFAULT)
+
+        # Local option to skip transformation.
+        if tropt == OPT_NONE:
+            return None
+
+        if transformations is None:
+            if tropt != OPT_DEFAULT:
+                self.logger.warning('transform specified without cautodoc_transformations config.',
+                                    location=(env.docname, self.lineno))
+
+            return None
+
+        transform = transformations.get(tropt)
+
+        if transform is None:
+            if tropt != OPT_DEFAULT:
+                self.logger.warning(f'unknown transformation "{tropt}".',
+                                    location=(env.docname, self.lineno))
+            return None
+
+        return lambda comment: transform(comment, transform=tropt)
 
     def __parse(self, viewlist, filename):
         env = self.state.document.settings.env
@@ -124,6 +164,7 @@ def setup(app):
     app.require_sphinx('3.0')
     app.add_config_value('cautodoc_root', app.confdir, 'env')
     app.add_config_value('cautodoc_compat', None, 'env')
+    app.add_config_value('cautodoc_transformations', None, 'env')
     app.add_config_value('cautodoc_clang', None, 'env')
     app.add_directive_to_domain('c', 'autodoc', CAutoDocDirective)
 
--- a/test/example-75-transform.c
+++ b/test/example-75-transform.c
@@ -1,0 +1,15 @@
+/**
+ * Custom comment transformations.
+ *
+ * Transformations require ``cautodoc_transformations`` configuration in
+ * ``conf.py``. In this example, Napoleon is used to interpret another
+ * documentation comment format.
+ *
+ * Args:
+ *     foo: This is foo.
+ *     bar: This is bar.
+ *
+ * Return:
+ *     Status.
+ */
+int napoleon(int foo, char *bar);
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -57,6 +57,7 @@ except ImportError:
     extensions.append('automock')
 
 from sphinx.ext.napoleon import docstring, Config
+from hawkmoth.util import doccompat
 
 def napoleon_transform(comment, **options):
     config = Config(napoleon_use_rtype=False)
@@ -64,6 +65,7 @@ def napoleon_transform(comment, **options):
 
 cautodoc_transformations = {
     'napoleon': napoleon_transform,
+    'javadoc-liberal': doccompat.javadoc_liberal,
 }
 
 # Add any paths that contain templates here, relative to this directory.
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -56,6 +56,16 @@ except ImportError:
     # into the documentation instead of generating.
     extensions.append('automock')
 
+from sphinx.ext.napoleon import docstring, Config
+
+def napoleon_transform(comment, **options):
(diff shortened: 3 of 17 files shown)


hawkmoth/__init__.py
```diff
@@ -62,15 +63,54 @@ def __display_parser_diagnostics(self, errors):
...
+        tropt = self.options.get('transform', OPT_DEFAULT)
+
+        # Local option to skip transformation.
+        if tropt == OPT_NONE:
+            return None
+
+        if transformations is None:
+            if tropt != OPT_DEFAULT:
+                self.logger.warning('transform specified without cautodoc_transformations config.',
+                                    location=(env.docname, self.lineno))
```
BrunoMSantos, line 100: If the user specifies `none` it won't trigger this check, although that could arguably be the intended behaviour. Still, overall, I think this function is not particularly clear, and it handles a couple of cases in ways that may be surprising:
- The fact that `default` defaults to "no transformation" is neither documented nor obvious when reading this function.
- If the user specifies a `none` option in `cautodoc_transformations`, that is not flagged as an error, but this code will ignore it all the same which could be really puzzling for the user. Still on this topic, `none` may (mere supposition) also carry different meanings in different languages as a word or acronym, so its absence should either be enforced or we need to consider it as a potential valid key.

Suggestion (totally untested):
```python
    def __get_transform(self):
        env = self.state.document.settings.env

        # Special case values for transform.
        OPT_NONE = 'none'
        OPT_DEFAULT = 'default'

        # Handle deprecated compat. To be removed.
        transform = self.__get_compat_transform()
        if transform is not None:
            return transform

        transformations = env.config.cautodoc_transformations
        tropt = self.options.get('transform', OPT_DEFAULT)

        if transformations is None and tropt != OPT_DEFAULT:
            self.logger.warning('transform specified without cautodoc_transformations config.',
                                location=(env.docname, self.lineno))
            return None

        # Use an identity transformation for the none and default
        # transformations by default.
        if transformations.get(OPT_NONE) is None:
            transformations[OPT_NONE] = lambda comment: comment
        if transformations.get(OPT_DEFAULT) is None:
            transformations[OPT_DEFAULT] = lambda comment: comment

        transform = transformations.get(tropt)
        if transform is None:
            self.logger.warning(f'unknown transformation "{tropt}".',
                                location=(env.docname, self.lineno))
            return None

        return lambda comment: transform(comment, transform=tropt)
```

hawkmoth/__init__.py
```diff
@@ -62,15 +63,54 @@ def __display_parser_diagnostics(self, errors):
...
+
+        transform = transformations.get(tropt)
+
+        if transform is None:
+            if tropt != OPT_DEFAULT:
+                self.logger.warning(f'unknown transformation "{tropt}".',
+                                    location=(env.docname, self.lineno))
+            return None
+
+        return lambda comment: transform(comment, transform=tropt)
```
BrunoMSantos, line 112: Hmm, I think we discussed this in a different iteration, maybe not though. We derive the `transform` function from `tropt` and then pass `tropt` as an argument to it again. This feels redundant and makes me wonder if some extraneous look-up is being done elsewhere.

Optionally passing extra options to the transformation function may be interesting although not that useful I think (one could simply define more entries in the transformation table), but the key shouldn't matter at all. I suspect this is a remnant of how `doccompat.convert()` works.

If you want to have additional options, I would suggest something like:
```python
cautodoc_transformations = {
    'key': { 'fn': function,
             'opt': { ... } },
    ...
}
```

test/example-75-transform.c
```diff
@@ -0,0 +1,15 @@
...
+ * documentation comment format.
+ *
+ * Args:
+ *     foo: This is foo.
+ *     bar: This is bar.
+ *
+ * Return:
+ *     Status.
+ */
+int napoleon(int foo, char *bar);
```
BrunoMSantos, line 15: Congrats, it's really cool how this all ties together!

BrunoMSantos (changes requested):
Great work! Had a couple of comments that I think are actionable though.

hawkmoth/__init__.py
jnikula: I know, I was struggling with this a bit, trying to make it least surprising.

It's intentional that `none` and `default` don't flag an error if `cautodoc_transformations` is not specified. They are the special cases.

The downside with adding `none` and `default` to transformation is that then all comments go through the no-op lambda, while it could be avoided by returning `None`.

hawkmoth/__init__.py
jnikula: Ditching the `transform` keyword option would be easy enough. Just pass it to `doccompat.convert()` as long as we have it.

What about the whole `**options` for transform functions? I'm thinking it's an easy way to pass useful information in the future. Do you suggest to remove just the one option that seems redundant, or the whole kwarg?

hawkmoth/__init__.py
jnikula: I'm trying to say, I'm not completely happy with my original function, I agree with your criticism, but I'm also not happy with your suggestion. Need to think about this.

hawkmoth/__init__.py
BrunoMSantos: Indeed, I was only referring to the tautological parameter which let's a function know that we called it. The `**options` will never be _needed_ I believe, but I like it except perhaps for how they can be defined. E.g. the suggestion I give above is too verbose in the general case (no options).

By the way, the reason I believe the options will never be a hard requirement is that one could simply (e.g.) create lambdas which captured the options we wanted and then place the lambdas (now without any extra options) on the table to no great loss.

jnikula: I was thinking of passing e.g. the Sphinx app or relevant Clang objects in the future, for more advanced stuff. Can't handle that with lambdas alone. But perhaps it's just too complicated for what Hawkmoth aims to be?

Should we just ditch the `**options` altogether? We can bring it back in the future if we need it. It's not like this is all carved in stone.

BrunoMSantos: Oh, so those options would be injected inside hawkmoth rather than through the table. In that case I'm more conflicted. On one hand if we remove it, users will have to update their transformation functions once we add it back in, if we leave it in, they may rightfully ask what it is for when they have no access to it by design.

I guess it depends on how likely you see that happening...

hawkmoth/__init__.py
jnikula: Okay, so here's a suggestion to remove `none` as a special case: allow `None` as a no-op "function" in `cautodoc_transformations`:

```
cautodoc_transformations = {
    'foo' = None,
}
```

Using `foo` in the directive transform option would explicitly do no transformation. This 1) skips the lambda when no transformation is needed, and 2) removes the special case key.

jnikula: Is it bad form to use `None` as a dict key for the default case? You wouldn't be able to explicitly choose the default case (because you can only have a string in the directive option), but I don't think that's a big deal.

I guess alternatively the empty string as a key could work be made to work too?

jnikula: This is what it would look like (the options part is not resolved yet). Unless using `None` as key is badly frowned on, I'd roll with this.

```python
    def __get_transform(self):
        env = self.state.document.settings.env

        # Handle deprecated compat. To be removed.
        transform = self.__get_compat_transform()
        if transform is not None:
            return transform

        transformations = env.config.cautodoc_transformations
        tropt = self.options.get('transform')

        if transformations is None:
            if tropt is not None:
                self.logger.warning('transform specified without cautodoc_transformations config.',
                                    location=(env.docname, self.lineno))

            return None

        # Note: None is a valid key for default.
        if tropt not in transformations:
            if tropt is not None:
                self.logger.warning(f'unknown transformation "{tropt}".',
                                    location=(env.docname, self.lineno))
            return None

        # Note: None is a valid value for no transformation.
        transform = transformations.get(tropt)
        if transform is None:
            return None

        return lambda comment: transform(comment, transform=tropt)

```

hawkmoth/__init__.py
jnikula: I'm leaning towards dropping the `**options` out for now. If we find out it'll be useful, we can add it back. If we don't want to break backward compatibility, I think we can use inspect to figure out what we need to pass. https://docs.python.org/3/library/inspect.html#inspect.signature

Of course we could also create complex objects etc. for this, but I'd like to start simple.

hawkmoth/__init__.py
jnikula: #50 is what everything would look like with this change.

jnikula: @BrunoMSantos I hate to ping for feedback, especially as I haven't been exactly fast with my responses either, but do you think you'll be able to spare a moment to look at this? I'd really like to get this out the door and do a release.

jnikula: Obviously I could just push, but I value your feedback!

BrunoMSantos: Yeah, I know I'm falling behind again, real life gets in the way sometimes. I'll have a look today after work then.

jnikula:
Superseded by #50

PR #50, opened 2021-05-15
Transformations v3

jnikula:
The discussion is still ongoing in #49, and I guess we should finish it there. This pull request reflects my last comments:

- Remove the special keywords ``none`` and ``default`` altogether, and use ``None`` value instead. (Use of ``None`` may be debatable, but let's finish the discussion in #49.)
- Remove ``**options`` from transformations for simplicity.

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -38,6 +38,7 @@ class CAutoDocDirective(Directive):
     final_argument_whitespace = True
 
     option_spec = {
+        'transform': directives.unchanged_required,
         'compat': directives.unchanged_required,
         'clang': directives.unchanged_required,
     }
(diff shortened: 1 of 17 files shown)


hawkmoth/__init__.py
```diff
@@ -71,15 +72,46 @@ def __get_clang_args(self):
...
+
+        # Note: None is a valid key for default.
+        if tropt not in transformations:
+            if tropt is not None:
+                self.logger.warning(f'unknown transformation "{tropt}".',
+                                    location=(env.docname, self.lineno))
+            return None
+
+        # Note: None is a valid value for no transformation.
+        return transformations.get(tropt)
```
BrunoMSantos, line 113: This turned out to be a lot cleaner! +1

jnikula:
@BrunoMSantos Many thanks for your review, much appreciated!

PR #54, opened 2021-09-08
Testing improvements

jnikula:
While refactoring parser code I ended up breaking the CLI tool a number of times. So let's start off with adding a test for the CLI. Additionally, add testing using the Sphinx html output, as well as some other minor fixes and improvements.

PR #53, opened 2021-06-30
Autodoc draft

jnikula:
This is a draft for implementing #51.
The big missing items are tests and incorporating the struct/union/enum member documentation recursively too.
There's some basic documentation included, but it's simply:
```
.. c:autofunction:: file.c function_name
```
and similarly for other types of symbols. The `c:auto<X>::` directive names map directly to the [Sphinx C Domain](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-c-domain) `c:<X>::` directives.
I'm still a bit unsure about the directive arguments (mainly about creating something that's not easy to extend). I'll keep pondering.

PR #55, opened 2021-09-08
Comment classes

jnikula:
This is based on #54 and includes everything there, but this gets a little more involved in the parser guts.

To properly implement the new autodoc features in #53, we'll need the parser to return the comments in a hierarchy instead of a flat list. Add classes to encapsulate the parsed comments in a tree. As an added bonus, we're able to postpone transformations until after the parse. This is good for simplicity of the parser (it really doesn't need to know) as well as caching the parsed results in #53.

This is all under the hood refactoring, with no user visible changes.


jnikula:
@BrunoMSantos I've actually already got local changes that superseed this. And the new stuff is based on top of #58 that I just sent out. I'll update this one too.

jnikula:
Superseeded by #59.

PR #57, opened 2021-09-09
Venv updates

jnikula:
Make the requirements.txt and venv more geared towards bleeding edge development, and nothing else.


jnikula:
Pushed.

PR #59, opened 2021-09-09
Comment classes v2

jnikula:
This is v2 of #55 and includes the prep work from #58.


BrunoMSantos (approved):
So I left some loose comments, but they're just small nitpicks. The whole thing is really nice and a massive improvement. Not to mention a huge step towards the additional directives and so on.

jnikula:
Superseeded by #60

PR #60, opened 2021-09-11
Comment classes v3

jnikula:
This incorporates #58 and #59 with review comments from @BrunoMSantos addressed:

- Run `update-examples` directly using the shebang instead of with interpreter
- Switched from json to yaml for test options (this is the only functional change here)
- Added _ prefix to private members
- Renamed the Comment classes to Docstring classes. I opted for the longer Docstring over Docstr. I hope this is a good compromise.
- Also renamed EnumVal to Enumerator and Func to Function in the Docstring classes, to better match the directive names.
- Renamed `get_rst` method to `get_docstring`, also opting for the longer name.
- Renamed `comment.py` to `docstring.py`.
- Some local variable renames
- Commit message changes to reflect the above

Additionally, at the end, decided to move the rest of the helpers from `hawkmoth.util.docstr` to `hawkmoth.docstring`, as I thought that's a good fit now.


jnikula:
For convenience, these two commands will give an overview of what's changed since v2 (git diff is not bright enough to understand the file rename, so have to do that manually):

```
git diff origin/comment-classes-v2 origin/comment-classes-v3 -- . :^hawkmoth/comment.py :^hawkmoth/docstring.py
git diff origin/comment-classes-v2:hawkmoth/comment.py origin/comment-classes-v3:hawkmoth/docstring.py
```

BrunoMSantos (commented):
I found one last issue which sadly escaped me last time. Not a big deal, but I think it should be fixed. Even then, this is great, don't think a new review round is required.

BrunoMSantos:
> (git diff is not bright enough to understand the file rename, so have to do that manually):

`git diff -M20%` or lower will also do the trick in this (fairly benign) rename case ;)

jnikula:
Merged.

jnikula:
> Hmm, sorry I missed this before, but `get_docstring` still references this member and the missing `_indent`. Sure, all derivatives add it back, so there is never a runtime error, but it feels wrong all the same.
> 
> Listing those in the constructor would mess up the neat initialization of each derivative class, so I'd suggest something like:
> 
> ```python
> class Docstring():
>     _indent = 0
>     _fmt = ''
>     # keep the constructor as is.
> ```
> 
> edit: `_indent` is from the next patch of course.

Thanks, fixed.

jnikula:
> That's very charitable of you joy

Nah, I deleted a file with your stuff in it, it's only fair!

Issue #51, opened 2021-06-16
Add support for pulling in docs for a single function/struct/enum


jekstrand:
I would really like to be able to write sphinx-style documentation where there is prose to tie everything together and then docs for structs, functions, etc. are pulled in where they make sense rather than wherever they happen to land in the file.  This seems especially important for C where you don't have strict class hierarchies grouping things together.  Unfortunately, hawkmoth can't do this, at least not today.

jnikula:
Thanks for the request. As said elsewhere, this is something I'd like to add. The little I've looked into it so far, there are some opens that I could use some early user feedback for. Otherwise I'll just do something that's easy for me. :)

1. For structs, unions and enums, is it enough granularity to be able to pull in the documentation for a struct/union and all its members together, and an enum and all its enumerator values together? They group nicely in C level, and I think would be more complicated to split up.
2. If you pick and choose documentation for specific C symbols, whose responsibility is it to pull in all the documentation from a file? The user? Or is the expectation to have an "all the rest" option or a warning for not pulling in everything?

jnikula:
For implementation details, some musing:

It's possible to extend the Sphinx Python autodoc extension for this. It does have a bunch of useful stuff. On the other hand, it's written for Python and some of that gets in the way. We've also repeatedly seen Sphinx just break backward compatibility for downstreams. The more intimately tied we are with Sphinx internals, the more likely we are to be caught by some incompatible upstream change.

Each C source file should only be parsed once (well, once per file + identical compiler flags), not each time there's a directive to parse something from that file. We need a way to store the parsed stuff across multiple directives, perhaps across multiple directives across multiple `.rst` documents.

@BrunoMSantos Any thoughts here? This is all related to https://gitlab.freedesktop.org/mesa/mesa/-/issues/4931 which could lead to a big time user for the project.

jekstrand:
For 1, I'm not sure what you're asking.  Are you looking for a comment at the top of the struct that has everything?  Or are you asking if I want to be able to pull in a single struct member?  I have no interest in pulling in a single member.  I just want to be able to say

    .. c:autostruct:: isl_struct
        :members:

And have it pull in everything.  A lot of doc solutions have a `:members:` thing you have to include if you want the full docs; otherwise you just get the over-all struct docs.  I don't know if you want to add that complexity but it might be good to require `:members:` even if you don't so we don't run into backwards compatibility issues if it ever is added.

[...]

PR #61, opened 2021-09-12
New auto* directives to pull in individual structs/functions/enums, etc.

jnikula:
This implements the basic feature request in #51.

While there may still be some rough edges in the implementation and documentation, I think the design is pretty good, and I think it's a good starting point to expand on. I can envision the following follow-up feature requests will come pretty soon:

- There needs to be a way to not repeat the source file name every time.
- There needs to be a way to grab, say, all structs in a source file without specifying all of them.
- After you've included documentation for some of the things you want to highlight, there needs to be a way to grab all the rest without specifying all of them.

They are not all that difficult to implement on top, just need to get the directive syntax right. For now, it's simply

`.. c:autofunction filename.c function_name
`

for all of them.

@BrunoMSantos After getting the much appreciated review for the previous changes from you, this is how I reward you, with another big batch of changes. ;)

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -39,8 +39,41 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-10-macro.c
    
+
+
+
+
+Automacro
+---------
+
+Source
+~~~~~~
+
+.. literalinclude:: examples/example-11-automacro.c
+   :language: C
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:automacro:: examples/example-11-automacro.c DIE
+      
+
+Output
+~~~~~~
+
+.. c:namespace-push:: Automacro
+
+.. c:automacro:: examples/example-11-automacro.c DIE
+   
+
+.. c:namespace-pop::
+
 
 Variable
 --------
@@ -62,8 +95,41 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-20-variable.c
    
+
+
+
+
+Autovar
+-------
+
+Source
+~~~~~~
+
+.. literalinclude:: examples/example-21-autovar.c
+   :language: C
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:autovar:: examples/example-21-autovar.c meaning_of_life
+      
+
+Output
+~~~~~~
+
+.. c:namespace-push:: Autovar
+
+.. c:autovar:: examples/example-21-autovar.c meaning_of_life
+   
+
+.. c:namespace-pop::
+
 
 Typedef
 -------
@@ -85,8 +151,41 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-30-typedef.c
    
+
+
+
+
+Autotype
+--------
+
+Source
+~~~~~~
+
+.. literalinclude:: examples/example-31-autotype.c
+   :language: C
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:autotype:: examples/example-31-autotype.c list_data_t
+      
+
+Output
+~~~~~~
+
+.. c:namespace-push:: Autotype
+
+.. c:autotype:: examples/example-31-autotype.c list_data_t
+   
+
+.. c:namespace-pop::
+
 
 Enum
 ----
@@ -108,8 +207,41 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-40-enum.c
    
+
+
+
+
+Autoenum
+--------
+
+Source
+~~~~~~
+
+.. literalinclude:: examples/example-41-autoenum.c
+   :language: C
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:autoenum:: examples/example-41-autoenum.c mode
+      
+
+Output
+~~~~~~
+
+.. c:namespace-push:: Autoenum
+
+.. c:autoenum:: examples/example-41-autoenum.c mode
+   
+
+.. c:namespace-pop::
+
 
 Struct
 ------
@@ -131,8 +263,41 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-50-struct.c
    
+
+
+
+
+Autostruct
+----------
+
+Source
+~~~~~~
+
+.. literalinclude:: examples/example-51-autostruct.c
+   :language: C
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:autostruct:: examples/example-51-autostruct.c list
+      
+
+Output
+~~~~~~
+
+.. c:namespace-push:: Autostruct
+
+.. c:autostruct:: examples/example-51-autostruct.c list
+   
+
+.. c:namespace-pop::
+
 
 Function
 --------
@@ -154,8 +319,13 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-70-function.c
    
+
+
+
 
 Preprocessor
 ------------
@@ -177,8 +347,41 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-70-preprocessor.c
    :clang: -DDEEP_THOUGHT
+
+
+
+
+Autofunction
+------------
+
+Source
+~~~~~~
+
+.. literalinclude:: examples/example-71-autofunction.c
+   :language: C
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:autofunction:: examples/example-71-autofunction.c frob
+      
+
+Output
+~~~~~~
+
+.. c:namespace-push:: Autofunction
+
+.. c:autofunction:: examples/example-71-autofunction.c frob
+   
+
+.. c:namespace-pop::
+
 
 Transform
 ---------
@@ -200,8 +403,13 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-75-transform.c
    :transform: napoleon
+
+
+
 
 Compat
 ------
@@ -223,8 +431,13 @@ Directive
 Output
 ~~~~~~
 
+
+
 .. c:autodoc:: examples/example-80-compat.c
    :transform: javadoc-liberal
+
+
+
 
 Generic
 -------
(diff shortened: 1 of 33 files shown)


BrunoMSantos:
Oh wow. I thought I'd squeeze in another review, but I can't do this today :/ I'll have more time later this week for sure though and I really want to review this one!

So far I had only a cursory look and like it so far. I have a few comments / questions, but I better leave them until I go over it in more detail. For now I will only comment on the API as that is something I had been doing differently in my own attempt a couple of months back:

Instead of
```
.. c:autofunction filename.c function_name 
```
Something a bit more verbose, sure, but also more 'Sphynxy' and that would better fit the future case where the filename becomes optional:
```
.. c:autofunction function_name
   :file: filename.c
```

jnikula:
Hey, no pressure on the review! (Though obviously I do want to get this merged. :)

The directive interface is the important one, isn't it? We can change the implementation at will, but changing the directive arguments and options is not going to go down well with the users.

BrunoMSantos:
That's a valid concern, but the old directive would remain the same. The way I see it, the file is a positional argument in the old directive, but would be an optional one in the new directives.

The only breaking change I would also argue for would be renaming the old directive as `automodule` to match up with Python's autodoc. Maybe leaving an alias for backwards compatibility? Anyway, I was hoping not to touch on that just yet ahah.

jnikula:
My point is, whatever new directive syntax we introduce now, we'll be stuck with, or we'll upset users.

jnikula:
Ok, I've thought about this, and I think the main argument against filename as directive argument (as in this pull request) is the use of a "filename context", i.e. filename to use set by a preceding directive to avoid repeating it. The directive option is a more natural fit for that.

Additionally, I'm thinking a `c:automodule` directive would be the natural way to set the filename context.

Here's my proposal:

Get all the top level comments from a file, and set the filename context for the
subsequent c:auto* directives:

```
.. c:automodule:: filename.c
```

Get all the documentation from a file, recursively, and set the filename context
for the subsequent c:auto* directives:

```
.. c:automodule:: filename.c
   :members:
```

Get all the top level comments from a file, as well as the documentation for the
named symbols, recursively, and set the filename context for the subsequent
c:auto* directives:

```
.. c:automodule:: filename.c
   :members: foo, bar
```

Set the filename context for the subsequent c:auto* directives ("no-text" or
"no-doc" or what?)

```
.. c:automodule:: filename.c
   :no-text:
```

Get the struct documentation from a named file, overriding the filename context,
if any, for this directive only:

```
.. c:autostruct:: name
   :file: filename.c
```

Get the struct documentation from the file set by preceding c:automodule
directive (*no* member documentation):

```
.. c:autostruct:: name
```

Get the struct document from the file set by preceding c:automodule directive,
with all members documented recursively:

```
.. c:autostruct:: name
   :members:
```

Get the struct document from the file set by preceding c:automodule directive,
with named members documented recursively:

```
.. c:autostruct:: name
   :members: foo, bar
```

Combined:

```
.. c:autostruct:: name
   :file: filename.c
   :members: foo, bar
```

Similarly for other c:auto* directives, except :members: is special for struct,
union and enum.

Potential *future* addition, multiple symbols at once:

```
.. c:autostruct:: name1 name2 or_even_glob_pattern*
```

Potential *future* addition, all symbols of said type:

```
.. c:autostruct::
```

Potential *future* addition, grouping at symbol level:

```
.. c:autostruct::
   :group: foo
```

Potential *future* addition, grouping at module level:

```
.. c:automodule:: filename.c
   :group: foo
```

The idea is that typically you'd want to grab the top level comments from a file and then document the contents, so a natural flow would be:

```
.. c:automodule:: filename.c

.. c:autofunction:: some_function

.. c:autoenum:: some_enum
```

And so on.

jnikula:
I don't think it's complicated to make the changes, but I'd like to reach an agreement before that. And, perhaps, get some review comments about the code also (apart from stuff that would change).

BrunoMSantos:
Ah, my bad. I took your original reply as disagreement. Yeah, I agree.

Regarding your proposal, that's pretty close to what I had in mind, except I don't think one directive should have any effect on any subsequent one. So `automodule` being a file selector for subsequent directives feels wrong to me, and likely to catch some users off guard as they (e.g.) move documentation blocks around.

I was holding off on this next section as this branch may already be compatible with what I have in mind regarding caching, but I think it may be good to discuss it now to really see where the API is going.

My idea was that we ultimately need a cache to solve the performance issue regardless of what we do to overcome the mandatory filename selection. So might as well assume a solution that involves a cache from the beginning and which can get both things done.

So, eventually, I would have an internal global flag stating whether a cache had been built or not and, if not, _any_ directive would trigger a rebuild, unless given some `:no-indexing:` option. This would mean going through the search path looking for source files and building a full cache of all of them, which I think is quite reasonable in principle. Possibly we could have filters in the `conf` file to exclude files or folders from the search path and so on or an option to disable indexing altogether.

Afterwards, any call would simply read from the cache and possibly omit the filename provided that there were no clashes. If there were, there would be an informative error as we know exactly how many different versions of a symbol exist and the user would have to go and add the `:file:` option to disambiguate. The `:file:` option would also be required when using `:no-indexing:` as that would bypass the cache.


I actually have a partial implementation of this, but it was not 'working' yet and it's now massively out of date and possibly fundamentally incompatible with some newer developments, so take that as you may.

jnikula:
Any alternative suggestions to `c:automodule` selecting the file to use? A separate directive only for that? I thought `c:automodule` would fit the bill as that would take the filename as directive argument as opposed to option in the other directives. It would be the first thing to do when starting to document a file.

From a user POV, I think having to explicitly state the filename (either using a separate directive or the `:file:` option) is less surprising than it sometimes magically working implicitly. Besides, it's something that we can relax later on if we like.

I have the cache implemented on a per-document basis, i.e. directives in one rst document will lead to parsing each referenced C file only once. That's trivial and only a handful of lines of code. I presume several different rst documents referencing the same C file is less common. So I'm not yet sold on complicating the caching for what I believe is not a common scenario.

BrunoMSantos:
> Any alternative suggestions to `c:automodule` selecting the file to use? A separate directive only for that? I thought `c:automodule` would fit the bill as that would take the filename as directive argument as opposed to option in the other directives. It would be the first thing to do when starting to document a file. 

I did not account for that at al, but say we went your way. I'd rather have a specific directive like (e.g.) `c:cachemodule` or `c:loadmodule` that produces no documentation whatsoever. That would be both more explicit and more versatile.

> From a user POV, I think having to explicitly state the filename (either using a separate directive or the `:file:` option) is less surprising than it sometimes magically working implicitly. Besides, it's something that we can relax later on if we like.

Hmm, if that's the problem, I would sooner drop the optional omission of the file name altogether and keep the full cache :/

In fact, your proposal carries arguably worse implicit assumptions on how the user writes or has written the code and the documentation. You even seem to be well aware of those assumptions as you name them below, only not giving them the same importance perhaps. Especially considering huge code bases trying to move away from Doxygen, which is what is driving this in fact. After all, I still believe the original directive had everything one should ever want!

> I have the cache implemented on a per-document basis, i.e. directives in one rst document will lead to parsing each referenced C file only once. That's trivial and only a handful of lines of code. I presume several different rst documents referencing the same C file is less common. So I'm not yet sold on complicating the caching for what I believe is not a common scenario.

Regardless of whether directives affecting other directives is a good idea, I think the crux of the problem may actually lie here. Particularly in that last statement.

Judging from the discussions we had with our only prospective (heavy) users so far, I believe both projects would want to do just that, quite possibly at scale. Doxygen has primed them with multi-file groups, multiple groups per file and not quite the same concept of top level comments. Given the time it takes (e.g.) Mesa to run all of their sources through the compiler, even multiplying that by 2 would probably be too much. I would also expect that source size correlates positively with the desire to split things, which mean the biggest and more difficult to compile files would be parsed disproportionately more times than the smaller ones.

Anyway, I don't want to hold this too much on this point. But I wonder about the course of action if (when?) we find ourselves discussing how to improve performance if a Mesa-like project goes all in. Is there a way to fix that without affecting the suggested API or making it obtuse somehow?

I can always try to rebase / adapt what I had before and demo what the full cache would look like. I was kind of putting that off until all of this early work was merged, but maybe it's important enough to do it now.

BrunoMSantos:
Side note: I haven't checked how you handle 'top level' comments that appear in the middle of the file... That will be an interesting case to explore and one I haven't thought through yet. Are there API implications on that one?

jnikula:
I don't actually expect us to figure this all out up front, and I am also not sure prospective users would know either. I mean sure, we can make guesses, and the users can make guesses, but we don't really know until they start using it, and hit the problems.

With that in mind, I think one of the design goals should be to start small in a way that can be extended later without breaking everything.

Conclusions from that:

- I'm changing my mind, *not* tying `c:automodule` to specifying the file is probably a good idea. We don't even need to to add `c:automodule` yet, or at all. (But we can also do either or both later.)
- Since all the directives would support the `:file:` directive option (and yes, I also agree with using a directive option instead of argument for this), we also don't need to add a directive to specify the file right away. (But we can add that later.)
- Since `:file:` is an option, we don't have to require it in the future if we come up with other ways to figure the file out. Whether it is via a directive or a cache of all C files.
- Per rst document caching is there now, it's simple, but it does not prevent us from incrementally adding better caching, or even up front parsing of all files.

I think changing `CAutoSymbolDirective._get_filenames()` to use the directive option instead of the argument would be sufficient to 1) provide users something useful to try out, and 2) provide us with a good base to experiment further work on.

BrunoMSantos:
Alright, that makes a lot of sense.

By the way, old me asked:
>  [...] Is there a way to fix that without affecting the suggested API or making it obtuse somehow?

I meant it rhetorically, but turns out there's a good answer I think. If there is a separate directive for loading up a file, which we seem to agree on, and we later decide we don't need it, we can make it a no-op and deprecate it rather painlessly. So much for my own objections :smile:

jnikula:
> I think changing `CAutoSymbolDirective._get_filenames()` to use the directive option instead of the argument would be sufficient to 1) provide users something useful to try out, and 2) provide us with a good base to experiment further work on.

Probably also need to support `:members:` for structs, unions and enums from the start. I'm not a huge fan of having to specify `:members:` to get any member documentation, but that aligns with https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html. So there are three scenarios:

- No `:members:` option, ignore member documentation.
- Empty `:members:` option, recursively include all member documentation.
- `:members:` option with comma-separated values, e.g. `:members: foo, bar`, recursively include named member documentation.

The last one can be left as a future addition without breaking backward compatibility. But adding the distinction between the first two afterwards does break things.

jnikula:
I split out refactoring to #62 and I'm rewriting the implementation according to what was agreed above, so I'm closing this one.

PR #66, opened 2021-09-25
FIx comment marker and line prefix stripping

### LINKED ISSUE #64: Continuation lines in parameter descriptions are broken
I have a long parameter description that looks like this:

```
  :param param1: an initialized variable
  :param param2: A structure containing various feature data, as configured in
                  SomeFunction and supported by the input data
```

The second parameter description is broken into two lines. This generates a warning "WARNING: Field list ends without a blank line; unexpected unindent." and renders as if I had written

```
  :param param1: an initialized variable
  :param param2: A structure containing various feature data, as configured in

  SomeFunction and supported by the input data
```

Is there a workaround for this, maybe some way of marking the param2-line as continuing on the next line? I know that the above documentation works fine in Python.

jnikula:
I'm not entirely happy with the implementation, feels like there's a prettier solution to be found, but I just couldn't find it now.

Fixes #64.

### DIFF
--- a/hawkmoth/docstring.py
+++ b/hawkmoth/docstring.py
@@ -61,16 +61,20 @@ class Docstring():
 
     def _get_plain_comment(self):
         """Return plain comment with comment markers and line prefixes removed."""
-        comment = self._text
+        lines = self._text.splitlines()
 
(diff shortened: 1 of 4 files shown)


jnikula:
Superseeded by #67

PR #65, opened 2021-09-23
Autodoc v4

jnikula:
Updated version of #63

The functional changes to the main code itself are small, only addressing review comments. Most changes since the last version are for tests and documentation.

PR #67, opened 2021-09-26
Comment strip v2

jnikula:
Rebase of #66, with `comment-strip.yaml` now added as that's required since #65.


jnikula:
I think I'll want this fixed before the next release.

jnikula:
Merged. I didn't think there was anything particularly controversial here, and can be fixed later if anything comes up.

PR #70, opened 2021-10-13
Define tests via yaml

jnikula:
Following up on test restructuring, define test cases through having a .yaml file instead of a .c file.


jnikula:
Thanks for the review!

PR #72, opened 2021-10-17
Improve handling of anonymous enums, structs and unions

jnikula:
Arguably you should name the things that deserve documentation at C level, but Hawkmoth should not trip over such cases either. Currently, the Clang anonymous placeholder names such as "(anonymous at foo.c:1:1)" leak through. Fix this by generating valid identifiers and using the Sphinx @-prefixed naming convention for anonymous entities.

I am not entirely happy how this turned out. I think the split of responsibilities between `parser.py` and `docstring.py` is right, but it somehow feels hackish. For some reason, Clang also doesn't provide full support via `.is_anonymous()` and we have to resort to regex.

### DIFF
--- a/test/enum.c
+++ b/test/enum.c
@@ -11,3 +11,17 @@ enum foo {
 	 */
 	baz,
 };
+
+/**
+ * Anonymous enum.
+ */
+enum {
+	/**
+	 * Enumeration comment.
+	 */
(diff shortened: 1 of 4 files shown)


jnikula:
Pushed, thanks for the review.

PR #74, opened 2021-11-15
Fix anonymous type detection for newer Clang version

BrunoMSantos:
Clang 13 introduces a change to the anonymous type names forcing us to
update the regex to support both versions.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -177,7 +177,7 @@ def _anonymous_fixup(ttype, name):
     if name:
         return ttype, name
 
-    mo = re.match(r'(?a)^(?P<type>enum|struct|union) ([^:]+::)?\(anonymous at [^)]+\)$', ttype)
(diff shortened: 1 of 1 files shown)


jnikula:
Thanks, and sorry for the delay! This is what I used to test the change:

```
diff --git a/docker/test/Dockerfile b/docker/test/Dockerfile
index fbd362deacc6..0bbdb6e1c5ca 100644
--- a/docker/test/Dockerfile
+++ b/docker/test/Dockerfile
@@ -1,8 +1,8 @@
-FROM debian:bullseye
+FROM debian:bookworm
 
 RUN apt-get update
 RUN apt-get install -y \
-       python3-clang \
+       python3-clang-13 \
        python3-pip
 RUN apt-get clean
```

At some point I think we'll have to define the version combinations of {Sphinx, Clang, Python} we want to test. But I guess for now we can still go ad hoc and fix issues as they arise.

jnikula:
Oops, we both forgot to run the new `make check` target for flake8. I opted to push a fix directly: 6dae9be214a8 ("flake8: silence a E501 line too long with "noqa: E501"")

PR #75, opened 2021-11-23
[RFC] Full cache

BrunoMSantos:
I've been holding this in different forms for quite a long time. Most of the complication and rebase conflicts were actually on the test folder and honestly I wasn't happy with any of it anyway, so I'm skipping it for now. We can at least have a discussion about the actual caching behaviour, which amounts to very little code changes in fact.

Note that I have no idea how to cleanly do the unit testing for this (ideas are welcome), but I did test it in an ad hoc fashion and it seems to work well.

I can also say that the changes did not measurably impact performance on my machine for the existing test set. For what it's worth, it actually scored better by a few seconds (cumulative) over 100 runs of `make test`. Then again I wasn't expecting any difference either.

### DIFF
--- a/doc/directives.rst
+++ b/doc/directives.rst
@@ -74,8 +74,32 @@ adding the ``file`` option.
       :type: text
 
       The ``file`` option specifies to file to parse. The filename is
-      interpreted relative to the :data:`cautodoc_root` configuration
-      option. (For the time being, this option is mandatory.)
+      interpreted relative to the :data:`cautodoc_root` configuration option.
+      Each parsed file is cached for any following directive, but it is never
+      kept across reruns and it takes :rst:dir:`c:autodoc:clang` into account,
+      which means the file may still be parsed multiple times for a single
+      build.
+
+      This option may be omitted with some caveats:
+
+      1. It will be generally less performant than specifying it: when omitted,
+         a full cache will be built based on all `\*.[ch]` files within or
+         nested within :data:`cautodoc_root`. This cache is built file by file
+         and skipped for any file previously cached. Omitting it in combination
+         with bespoke per directive :rst:dir:`c:autodoc:clang` options is
+         particularly costly, potentially triggering the parsing of all files
+         for each and every directive that does so.
+
+      2. If omitted, the variable ``name`` will be matched against each and
+         every file that is automatically discovered regardless of how many
+         matches are found. This is done on purpose to make sure that we raise a
+         warning when a symbol clashes and the user can fix it by explicitly
+         specifying a file.
+
+      3. If multiple matches are found, they are all included in the
+         generated documentation in the cache build order, which is in part
+         defined by the user and otherwise undefined from Hawkmoth's
+         perspective.
 
    For example:
 
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -97,32 +97,11 @@ class CAutoBaseDirective(SphinxDirective):
         # Note: None is a valid value for no transformation.
         return transformations.get(tropt)
 
-    def __parse(self, filename):
-        clang_args = self.__get_clang_args()
-
-        # Cached parse results per rst document
-        parsed_files = self.env.temp_data.setdefault('cautodoc_parsed_files', {})
-
-        # The output depends on clang args
-        key = (filename, tuple(clang_args))
-
-        if key in parsed_files:
-            return parsed_files[key]
-
-        # Tell Sphinx about the dependency
-        self.env.note_dependency(filename)
-
-        docstrings, errors = parse(filename, clang_args=clang_args)
-
-        self.__display_parser_diagnostics(errors)
-
-        parsed_files[key] = docstrings
-
-        return docstrings
-
-    def __get_docstrings(self, viewlist, filename):
+    def __get_docstrings(self, cache, clang_args, viewlist, filename):
         transform = self.__get_transform()
-        root = self.__parse(filename)
+
+        key = (filename, tuple(clang_args))
+        root = cache.get(key, None)
 
         for docstrings in root.walk(recurse=False, filter_types=self._docstring_types,
                                     filter_names=self._get_names()):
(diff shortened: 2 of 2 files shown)


hawkmoth/__init__.py
```diff
@@ -171,7 +177,16 @@ def run(self):
...
             self.__build_cache(cache, clang_args, filename)
-            self.__get_docstrings(cache, clang_args, result, filename)
+            if self.__get_docstrings(cache, clang_args, result, filename):
+                matches.append(filename)
+
+        nmatches = len(matches)
+        if nmatches == 0:
+            self.logger.warning('No matching docstring.',
+                                location=(self.env.docname, self.lineno))
+        elif nmatches > 1:
```
BrunoMSantos, line 187: Bah, this needs to be `auto and ...` or it would trigger in the `cautodoc` directive when issued with globs or multiple files.

jnikula:
Interesting! I'm afraid I might not have time to look into it until the weekend.

BrunoMSantos:
> Interesting! I'm afraid I might not have time to look into it until the weekend.

No worries!

doc/directives.rst
```diff
@@ -74,8 +74,30 @@ adding the ``file`` option.
...
+         :data:`cautodoc_root` unless one was built by a previous directive
+         already. Omitting it in combination with bespoke per directive
+         :rst:dir:`c:autodoc:clang` options is particularly costly, potentially
+         triggering the parsing of all files for each and every directive.
+
+      2. If omitted, the variable ``name`` will be matched across the whole set
+         of cached entries that match the :rst:dir:`c:autodoc:clang` options.
+         Note that this includes any file outside the :data:`cautodoc_root`
+         search path that has been previously cached by a directive which
+         explicitly included it.
```
BrunoMSantos, line 96: Kept thinking about this. This last sentence is a lie in the current version... Only files in or nested `cautodoc_root` will ever be considered in this case, even if others were cached in between.

hawkmoth/__init__.py
```diff
@@ -162,21 +173,28 @@ class CAutoDocDirective(CAutoBaseDirective):
     optional_arguments = 100   # arbitrary limit
 
     def _get_filenames(self):
+        filenames = []
+
         for pattern in self.arguments:
-            filenames = glob.glob(self.env.config.cautodoc_root + '/' + pattern)
-            if len(filenames) == 0:
+            candidates = glob.glob(os.path.join(self.env.config.cautodoc_root, pattern))
```
jnikula, line 209: Fixes like using os.path.join() should be done and merged first.

hawkmoth/__init__.py
```diff
@@ -151,12 +130,48 @@ def _get_filenames(self):
         """
         raise NotImplementedError(self.__class__.__name__ + '._get_filenames')
 
+    def __get_all_filenames(self):
+        nfiles = 0
+
+        filenames = glob.glob(os.path.join(self.env.config.cautodoc_root, '*.[ch]'))
+        filenames += glob.glob(os.path.join(self.env.config.cautodoc_root, '**/*.[ch]'))
```
jnikula, line 142: If I understand this correctly, encountering a directive without filenames specified will scan all the `*.[ch]` files in source tree. For every directive.

The obvious improvement would be to scan all the files only once at extension setup. However, I still have really mixed feelings about on the one hand restricting to `*.[ch]` suffix and on the other hand scanning the entire source tree, and parsing all the files if there's just one directive without a filename specified.

Say, if you have a moderately big project, but only have a handful of files to document, you'd end up running clang on every file in the project across the entire project.

hawkmoth/__init__.py
```diff
@@ -151,12 +130,48 @@ def _get_filenames(self):
...
+            self.env.note_dependency(filename)
+
+            docstrings, errors = parse(filename, clang_args=clang_args)
+
+            self.__display_parser_diagnostics(errors)
+            cache[key] = docstrings
+
     def run(self):
+        clang_args = self.__get_clang_args()
+        cache = self.env.temp_data.setdefault('cautodoc_cache', {})
```
jnikula, line 167: So my limited understanding of Sphinx parallel build (`sphinx-build -j N` argument) is that `self.env` is not shared during document reading i.e. each document has its own copy, and the data gets merged at a later stage.

See documentation for `parallel_read_safe` under https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata as well as `env-merge-info` event documentation at https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-env-merge-info

I can't be certain without actually digging into Sphinx source and/or testing, but I think this implementation is still a per-document cache, not a global cache. I'd be happy to be proven wrong, but this was my conclusion also when I opted for the simple cache.

jnikula: Also see the todo example extension at https://www.sphinx-doc.org/en/master/development/tutorials/todo.html
I think the only ways to benefit from global cache is to either
1) Disable parallel build in the extension setup, so we can use the cache synchronously across the documents.
**or**
2) Switch to a three phase build where the first phase simply gathers the directives and the filenames to parse, the second phase is the `env-merge-info` merges them, and the third phase is the `doctree-resolved` event where stuff gets parsed and output generated.

BrunoMSantos: Interesting. I didn't test with parallel build indeed so I failed to notice this. I also assumed that part was fine because this is how you handled the single file cache, which apparently does not cache as much as I thought it did. Of course this needs to be looked into before going any further.

Thanks for the references. I'll see what can be done about this.

hawkmoth/__init__.py
BrunoMSantos: > If I understand this correctly, encountering a directive without filenames specified will scan all the `*.[ch]` files in source tree. For every directive.

No, we take the file names on every directive, which might be a bit wasteful, but it's just walking the file system here. Clang will only be called once per file (not considering the issue you raised with parallel builds).

That said, we can take the file list only once though for a small performance improvement indeed.

> The obvious improvement would be to scan all the files only once at extension setup.

At setup we don't know if there are any directives that do not list a file name. Essentially, this means that we will always have worst case parsing even if the user never uses the auto discovery explicitly, which would be against your last point and which I'm also against.

This may be worked around if we instead have to pass all the filenames we're interested in through the configuration file instead of a directory. Or maybe a white / black list of source files.

And as for the globing pattern(s), we could default to `*.[ch]` but make it user configurable.

> However, I still have really mixed feelings about on the one hand restricting to `*.[ch]` suffix and on the other hand scanning the entire source tree, and parsing all the files if there's just one directive without a filename specified.
> 
> Say, if you have a moderately big project, but only have a handful of files to document, you'd end up running clang on every file in the project across the entire project.

Hmm, I don't think we can have it both ways, but that's why the documentation should be very clear that omitting the file names will likely trade performance for convenience.

BrunoMSantos: > > Say, if you have a moderately big project, but only have a handful of files to document, you'd end up running clang on every file in the project across the entire project.
> 
> Hmm, I don't think we can have it both ways, but that's why the documentation should be very clear that omitting the file names will likely trade performance for convenience.

Well, I guess a possible compromise would be to stop parsing when we get a match. But if anyone is ever going to rely on auto discovery, they should be assured that there is no ambiguity in their docs. And returning the 1st hit might miss the fact that there are multiple symbols with the same name. And the likelihood of such duplicate symbol names only increase with project size.

hawkmoth/__init__.py
BrunoMSantos: True, I'll split it up. Thanks for the comments by the way.

BrunoMSantos:
Sorry, I didn't mean to close this, it was a confusion on my part with the automatic mirror from GitLab... still learning how it all fits together. Feel free to keep the discussion going of course.

I'll open a new PR once I sort out the parallel build issue at least.

BrunoMSantos:
Nothing to see here! I just pushed some of the smaller changes I had worked on ages ago so that they lived elsewhere other than my hard drive. I haven't touched this in ages and don't remember what state it is in. Judging from CI, not a good one...

Maybe we should just close it for the time being?

jnikula:
It's not like anything gets lost if we close this one.

PR #76, opened 2021-12-11
Clean ups

BrunoMSantos:
Split from the full-cache branch. These are all quite trivial, but they should be merged regardless of what the full cache branch ends up looking like.


jnikula:
Apologies for the long delay. Merged now, thanks.

PR #77, opened 2022-01-14
change clang _Bool to more readable bool

ceepu:
when declare with bool, clang auto change it to _Bool.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -170,6 +170,8 @@ def _decl_fixup(ttype, name):
 
     ttype, name = _function_pointer_fixup(ttype, name)
 
+    if ttype == '_Bool': ttype = 'bool'
+
     return ttype, name
 
 # name may be empty for typedefs and anonymous enums, structs and unions


jnikula:
Thanks for your contribution!

Looks like the root cause is that `<stdbool.h>` defines `bool` instead of typedeffing it. Only the preprocessor sees `bool`, and from the compiler's perspective, it's `_Bool`. Contrast with `<stdint.h>` which typedefs e.g. `uint32_t`.

The trouble with your fix is that it's very specific to this particular problem. What if the source code had `_Bool`? Are there other cases where we should use whatever is in the source (and what the preprocessor sees) instead of the underlying type? I.e. should we look at the tokenized input instead?

Finally, I should really finish my guide for contributing. For any changes like this, I expect a test update that ensures we don't break this in the future. It's also helpful in demonstrating the need for the change. Also, please write a meaningful commit message.

hawkmoth/parser.py
```diff
@@ -170,6 +170,8 @@ def _decl_fixup(ttype, name):
 
     ttype, name = _function_pointer_fixup(ttype, name)
 
+    if ttype == '_Bool': ttype = 'bool'
```
jnikula, line 173: I think it's useful to follow the style we have for other fixups. One function per fixup, and the comments for the function can desribe what it's about.

jnikula:
I opened #92 to track this, but I'm closing the merge request.

PR #78, opened 2022-01-14
bug fix for type fixup *char[]

ceepu:
the original code parse `f(char *a[])` as `f(char *[]a)`, this patch fixed it.


jnikula:
Clearly a bug. The fix seems to be fine, but again, please add a test case, as well as a more elaborate commit message. Also, this is about generic arrays of pointers, not specifically `char *`.

jnikula:
Superseeded by #82 which uses a regex and includes test cases.

Issue #69, opened 2021-10-12
Function pointer variables and members omit parameter names


jnikula:
Given this C source:

```
/**
 * Function pointer foo.
 *
 * :param bar: bar
 * :param baz: baz
 */
void (*foo)(int bar, int baz);
```
you'll get this rst:

```
.. c:var::  void (*foo)(int, int)

   Function pointer foo.

   :param bar: bar
   :param baz: baz
```

The parameter names have been omitted, making the parameter documentation less than helpful. The same holds for struct members. See if the names are available in the clang parse results and add them if possible.

jnikula:
Since the parameter names aren't part of the type, the only way to get at them seems to parsing via `cursor.get_tokens()`. It can get pretty complicated, but it's doable. Also, they are not required in C.

Sphinx does support adding the names e.g. `.. c:member::  int (*foo)(int x, int y)` and they render as you'd expect.

PR #86, opened 2022-10-18
Add parameter names to function pointer variables and members

robinbourianes-kalisio:
Hi there, here's an implementation tentative to address issue #69
It may not meet the project's standards but hopefully it's something I can improve with your guidance.

I removed the regex parsing of the function pointer type and instead rely on clang's AST to rebuild the function signature, with names if available.

- Parameter names (and types) can be found in the children of the function pointer decl cursor.
- The associated FUNCTION_PROTO is used to find out about variadic function
- While traversing Type nodes to lookup the FUNCTION_PROTO I also accumulate required bits to rebuild function parameter declaration (constant array sizes and pointer qualifiers)

I also added a few tests to cover new use cases, although i could add a few more.
Tests run ok, except a few ones involving anonymous structs, the id appended to the anonymous structs is different from what's expected.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -149,28 +149,63 @@ def _array_fixup(ttype, name):
 
     return ttype, name
 
-# If this is a function pointer, or an array of function pointers, the
-# name should be within the parenthesis as in (*name) or (*name[N]).
-def _function_pointer_fixup(ttype, name):
-    mo = re.match(r'(?P<begin>.+)\((?P<stars>\*+)(?P<qual>[a-zA-Z_ ]+)?(?P<brackets>\[[^]]*\])?\)(?P<end>.+)', ttype)  # noqa: E501
-    if mo is None:
+# Return the FUNCTIONPROTO type associated with a function pointer cursor,
+# or None if not found
+# It also accumulates qualifiers and constant array sizes while traversing
+def _traverse_to_function_proto(cursor):
+    stars_and_quals = ''
+    brackets = ''
+    elt_type = cursor.type
+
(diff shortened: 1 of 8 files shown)


jnikula:
Thanks for the contribution, looks interesting! I'm afraid I'm swamped at the moment, it may take a bit before I'll have a moment to look into this deeper.

At a glance, one thing that I think might be broken is that I want tests to pass for every commit, not just before/after each merge request. `git rebase -i origin --exec="make test"` or something like that. Ditto for `make check`.

Note to self 1, I really need to get my contributing guidelines merged.

Note to self 2, I really need to get github actions working to check all this automagically.

jnikula:
As a feature, this is definitely something I'd like to have! Thanks!
It's somewhat of a bummer that Sphinx renders parameter field lists differently within `.. c:function::` and `.. c:member::`. The former has a nicer formatting, while the latter is just a plain field list. Not sure if there's anything that could be done, and definitely not a blocker.
I'll still need to try to wrap my head around the parser bits. :)

robinbourianes-kalisio:
Sure, take your time, no problem. I'm also a bit swamped at the moment.
Yeah, I noticed that parameters were not as nice as in `.. c:function::` and i don't know Sphinx enough to understand if something can be done about that. One step at a time !

jnikula:
@robinbourianes-kalisio I didn't feel comfortable asking you to fix the extra stuff I had in mind, so I kind of took the great start you contributed, and ran with it: #93. I would not easily have figured this out myself, so this is really much appreciated. Thanks!

robinbourianes-kalisio:
That's fine by me, glad to contribute !

PR #91, opened 2022-10-23
Cleanup formatting

BrunoMSantos:
Looking over Robin's #86 I had a few ideas and started poking around only to notice that the parser and docstring modules still had a few blurry lines in their boundaries. This is a quick attempt at fixing those issues so that the parser has 0 effect in the output formatting. It also makes the formatting a bit more consistent overall.

This will conflict with #86 though it's not a big issue and I have local patches for it already if it matters.

### DIFF
--- a/hawkmoth/docstring.py
+++ b/hawkmoth/docstring.py
@@ -146,12 +146,17 @@ class Docstring():
 
         ttype = self._ttype
 
+        spacer = ''
+        if ttype and not (len(ttype) == 0 or ttype.endswith('*')):
+            spacer = ' '
+
         args = ''
         if self._args and len(self._args) > 0:
             arg_fmt = lambda t, n: f"{t}{'' if len(t) == 0 or t.endswith('*') else ' '}{n}"
             args = ', '.join([arg_fmt(t, n) for t, n in self._args])
 
-        rst = self._fmt.format(text=text, name=self._get_decl_name(), ttype=self._ttype, args=args)
+        rst = self._fmt.format(text=text, name=name, ttype=ttype,
+                               type_spacer=spacer, args=args)
 
         rst = Docstring._nest(rst, self._nest)
 
@@ -175,7 +180,7 @@ class TextDocstring(Docstring):
 
 class VarDocstring(Docstring):
     _indent = 1
-    _fmt = '\n.. c:var:: {ttype} {name}\n\n{text}\n'
+    _fmt = '\n.. c:var:: {ttype}{type_spacer}{name}\n\n{text}\n'
 
 class TypeDocstring(Docstring):
     _indent = 1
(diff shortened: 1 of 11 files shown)


hawkmoth/docstring.py
```diff
@@ -142,7 +142,14 @@ def get_docstring(self, transform=None):
 
         text = Docstring._nest(text, self._indent)
 
-        args = ', '.join(self._args) if self._args is not None else None
+        name = self._get_decl_name()
```
jnikula, line 145: I guess you meant to use name later, but didn't?

hawkmoth/docstring.py
```diff
@@ -142,7 +142,14 @@ def get_docstring(self, transform=None):
...
+
+        ttype = self._ttype
+
+        args = ''
```
jnikula, line 153: Feels like this should remain None if there are no args. If it "leaks" to `_fmt.format()` I think it'll be more obvious there's a bug somewhere?

jnikula: Ahah, used in the 2nd commit.

jnikula (commented):
Overall I like it. The whitespace stuff has annoyed me too. And I agree with the spacing `int *foo` over anything else.

Couple of nits, I could lean either way with them.

hawkmoth/docstring.py
jnikula: Sheesh github landed this comment in a silly place. This was related to `name = self._get_decl_name()`.

hawkmoth/docstring.py
BrunoMSantos: Ah, it was for consistency and readability, I wanted to have a smaller name to pass to the function. But when I decided the commits made more sense in the reverse order, I clearly messed up this part of the rebase.

I'll probably need to force push to resolve it.

hawkmoth/docstring.py
BrunoMSantos: So I may be misreading this sequence of comments, but I'm guessing the 1st comment is still about `args = ''`? If so, note that `self._args` may still be `None`, `args` on the other hand is the processed argument string we want to use in the final line and that has to be an actual string. Did I misunderstand anything?

By the way, there could be a point in handling `self._args` in the derived classes. I could create a method to abstract the argument stitching. Same for the type spacer thing. It felt a bit verbose in comparison, but might make it nicer overall.

jnikula: I meant that it's a (theoretical) functional change to set `args = ''`.  It was set to `None` before, but it's fine when `self._fmt` does not contain `{args}`.
Anyway, not a big deal, I think I'll just merge this as-is.
Thanks for doing this!

BrunoMSantos: Ah, I see now. Well, I didn't fix the `name` and `ttype` being defined one commit ahead, but alright.
Thanks for the review!

PR #93, opened 2022-11-08
Param names wip

jnikula:
Building on the foundations laid by @robinbourianes-kalisio in #86. Throw the array regexp decl fixup to the curb as well, and do everything using clang. Still a wip, the whitepace handling is ugly to say the least.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -172,7 +172,8 @@ def _decl_fixup(cursor):
         for c in cursor.get_children():
             if c.kind == CursorKind.PARM_DECL:
                 arg_ttype, arg_name = _decl_fixup(c)
-                args.append(f'{arg_ttype}')
+                spacer = '' if not arg_name or arg_ttype.endswith('*') else ' '
+                args.append(f'{arg_ttype}{spacer}{arg_name}' if arg_name else arg_ttype)
         if cursor_type.is_function_variadic():
             args.append('...')
         if not args:
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -139,41 +139,57 @@ def _get_macro_args(cursor):
 
 # If this is an array, the dimensions should be applied to the name, not
 # the type.
-def _array_fixup(ttype, name):
-    mo = re.match(r'(?P<ttype>[^][]*)(?P<dims>(\[[^]]*\])+)$', ttype)
-    if mo is None:
-        return ttype, name
-
-    ttype = mo.group('ttype').strip()
-    name = name + mo.group('dims')
-
-    return ttype, name
 
 # If this is a function pointer, or an array of function pointers, the
 # name should be within the parenthesis as in (*name) or (*name[N]).
-def _function_pointer_fixup(ttype, name):
-    mo = re.match(r'(?P<begin>.+)\((?P<stars>\*+)(?P<qual>[a-zA-Z_ ]+)?(?P<brackets>\[[^]]*\])?\)(?P<end>.+)', ttype)  # noqa: E501
-    if mo is None:
-        return ttype, name
-
-    begin = mo.group('begin')
-    stars = mo.group('stars')
-    qual = mo.group('qual') + ' ' if mo.group('qual') is not None else ''
-    brackets = mo.group('brackets') if mo.group('brackets') is not None else ''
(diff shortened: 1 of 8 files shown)


BrunoMSantos:
Nice one! This is a massive improvement on its own and curiously, I was working on this and stopped right about where you are now too. What I think is missing though is that it doesn't leave the final output composition to `docstring` as I had hoped, which I guess is what you feel as well.

I didn't think this through enough, but my idea was roughly to extend `ttype` to capture function types more generally. Something like `('star/quals', 'return type', (('arg type', 'arg name'), ...))`. Then in `docstring` we could leverage the same argument concatenation (which is already there to parse `args`) and abstract the generation of a function pointer signature which is then used in `Var`, `Member` and `Type` sub classes for these complex 'types'.

I would also object to the name `decl_fixup` at that point because we're not so much fixing as just parsing further elements. Well, we're still 'fixing' the array stuff into the name I guess.

The counter argument to all of this is that at some point we're recreating the Clang's AST...

hawkmoth/parser.py
```diff
@@ -139,41 +139,57 @@ def _get_macro_args(cursor):
...
+    if cursor_type.kind == TypeKind.FUNCTIONPROTO:
+        args = []
+        for c in cursor.get_children():
+            if c.kind == CursorKind.PARM_DECL:
+                arg_ttype, arg_name = _decl_fixup(c)
+                args.append(f'{arg_ttype}')
+        if cursor_type.is_function_variadic():
+            args.append('...')
+        if not args:
+            args.append('void')
```
BrunoMSantos, line 180: There's no equivalent thing for a normal function, we just leave it empty. I think we should add the void there as well for consistency. I haven't double checked, but I think we're missing a test case or two by the way.

jnikula: Observed the same.

jnikula:
> Nice one! This is a massive improvement on its own and curiously, I was working on this and stopped right about where you are now too. What I think is missing though is that it doesn't leave the final output composition to `docstring` as I had hoped, which I guess is what you feel as well.

I started off reviewing #86 which basically meant I had to wrap my head around the whole parsing thing, and realized the array fixup could be done at the same time. I hope we didn't do too much duplicated work, although for me this was also about learning the stuff so I could review it in the first place, so no big deal I guess.

> I didn't think this through enough, but my idea was roughly to extend `ttype` to capture function types more generally. Something like `('star/quals', 'return type', (('arg type', 'arg name'), ...))`. Then in `docstring` we could leverage the same argument concatenation (which is already there to parse `args`) and abstract the generation of a function pointer signature which is then used in `Var`, `Member` and `Type` sub classes for these complex 'types'.
> 
> I would also object to the name `decl_fixup` at that point because we're not so much fixing as just parsing further elements. Well, we're still 'fixing' the array stuff into the name I guess.
> 
> The counter argument to all of this is that at some point we're recreating the Clang's AST...

I guess the main point is, given a `Cursor`, how do you format it suitable for Sphinx? For now, we've managed to separate the Clang and Sphinx parts by having the `Docstring` class structure as a kind of intermediate representation. We pass pre-parsed stuff to `Docstring`.

The alternative would be to make `Docstring` aware of Clang, and only pass the relevant `Cursor` to each `*Docstring` init, and have them sort stuff out internally. It might not be all that bad, really.

BrunoMSantos:
> I hope we didn't do too much duplicated work, although for me this was also about learning the stuff so I could review it in the first place, so no big deal I guess.

Nah, besides only the array stuff was really new so there was no big cost in duplicating work.

> > The counter argument to all of this is that at some point we're recreating the Clang's AST...
> 
> [...] 
> The alternative would be to make `Docstring` aware of Clang, and only pass the relevant `Cursor` to each `*Docstring` init, and have them sort stuff out internally. It might not be all that bad, really.

Hmm, maybe? Any specific gripe with the proposed idea of extending what we interpret as `ttype` though or more generally work in a simplified intermediate representation for our purposes?

Passing the cursor may lead to duplication of the AST walk code by the way. Something I'd like to avoid. However maybe we can abstract the AST walk itself. Come to think of it there may be some very idiomatic way of doing just that. 

Such a rework would never be part of this PR, of course. In that case we might well merge it as is for the time being.

jnikula:
Superseded by #94

PR #94, opened 2022-11-17
Param names

jnikula:
Updated version of #93.

I'm still not entirely happy about how all the spacing happens, but I think it's passable. Fixed void functions while at it, and added some more tests.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -302,6 +302,8 @@ def _recursive_parse(comments, cursor, nest):
 
             if cursor.type.is_function_variadic():
                 args.extend([('', '...')])
+            if len(args) == 0:
+                args.extend([('', 'void')])
 
         ttype = cursor.result_type.spelling
 
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -167,7 +167,10 @@ def _function_pointer_fixup(ttype, name):
 
     return ttype, name
 
-def _decl_fixup(ttype, name):
+def _decl_fixup(cursor):
+    ttype = cursor.type.spelling
+    name = cursor.spelling
+
     ttype, name = _array_fixup(ttype, name)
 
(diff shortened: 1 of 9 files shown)


hawkmoth/parser.py
```diff
@@ -139,41 +139,60 @@ def _get_macro_args(cursor):
...
-    return ttype, name
+    if cursor_type.kind == TypeKind.FUNCTIONPROTO:
+        args = []
+        for c in cursor.get_children():
+            if c.kind == CursorKind.PARM_DECL:
+                arg_ttype, arg_name = _decl_fixup(c)
+                args.append(f'{arg_ttype}')
+        if cursor_type.is_function_variadic():
+            args.append('...')
+        if not args:
```
BrunoMSantos, line 182: Minor nitpick: maybe `len(args) == 0` is more expressive? Same for a later commit.

test/function.c
```diff
@@ -4,14 +4,14 @@
...
 /**
- * Void parameter.
+ * No parameters.
  */
-int fooing(void);
+int no_parameters(void);
 
 /**
- * Empty parameters.
+ * Empty parameter list.
```
BrunoMSantos, line 12: 'unknown' might be more descriptive, no?

BrunoMSantos (approved):
Couple of minor nitpicks, but I can go either way. Otherwise I think it's good to go notwithstanding what we discussed in the draft version.
It's also a great clean-up of course, congrats.

test/function.c
jnikula: The standard talks about "empty list".

jnikula:
> Couple of minor nitpicks, but I can go either way. Otherwise I think it's good to go notwithstanding what we discussed in the draft version. It's also a great clean-up of course, congrats.

Thanks for the review! Fixed the code change, left the test as-is, and merged.

PR #95, opened 2022-11-17
Troubleshooting

jnikula:
Install `hawkmoth` as a console script, and document some troubleshooting tips (incl. using said script).


BrunoMSantos:
Nice one! LGTM

PR #97, opened 2022-12-01
Diagnostics

### LINKED ISSUE #80: Fix diagnostics when hitting unhandled cursor
As noted in #79, having a documentation comment before an include directive leads to an unhandled cursor error. The runtime diagnostics should be improved, instead of adding the error message in the documentation. Alternatively, it should directly fall back to a top level comment.

jnikula:
Fix #80 with some cleanups while at it

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -8,6 +8,7 @@ Sphinx C Domain autodoc directive extension.
 """
 
 import glob
+import logging as log
 import os
 
 from docutils import nodes, statemachine
@@ -17,7 +18,7 @@ from sphinx.util.nodes import nested_parse_with_titles
 from sphinx.util.docutils import switch_source_input, SphinxDirective
 from sphinx.util import logging
 
-from hawkmoth.parser import parse, ErrorLevel
+from hawkmoth.parser import parse
 from hawkmoth.util import doccompat, strutil
 from hawkmoth import docstring
 
@@ -37,16 +38,19 @@ class CAutoBaseDirective(SphinxDirective):
 
     _docstring_types = None
 
-    # Map verbosity levels to logger levels.
-    _log_lvl = {ErrorLevel.ERROR: logging.LEVEL_NAMES['ERROR'],
-                ErrorLevel.WARNING: logging.LEVEL_NAMES['WARNING'],
-                ErrorLevel.INFO: logging.LEVEL_NAMES['INFO'],
-                ErrorLevel.DEBUG: logging.LEVEL_NAMES['DEBUG']}
+    # Map parser diagnostic levels to verbosity levels.
+    def __parser_verbosity(self, level):
+        if level >= log.ERROR:
+            return 0
+        elif level >= log.WARNING:
+            return 1
+        else:
+            return 3
 
     def __display_parser_diagnostics(self, errors):
         for error in errors:
-            if error.level.value <= self.env.app.verbosity:
-                self.logger.log(self._log_lvl[error.level], error.get_message(),
+            if self.__parser_verbosity(error.level) <= self.env.app.verbosity:
+                self.logger.log(error.level, error.get_message(),
                                 location=(self.env.docname, self.lineno))
 
     def __get_clang_args(self):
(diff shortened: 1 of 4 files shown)


hawkmoth/__init__.py
```diff
@@ -8,6 +8,7 @@
 """
 
 import glob
+import logging as log
```
BrunoMSantos, line 11: So far I'm enjoying these patches, but I have serious doubts about this one. As far as I can see, you don't save nearly enough verbosity to justify the obfuscation of renaming one of the core modules. So I'd rather do `from sphinx.util import logging as log` all other considerations aside.

However, I think using the levels directly from `logging` is also wrong technically, even if it happens to produce the right outcome _today_. Note that `sphinx.util.logging` is clearly based on Python's `logging` (we can even find it inside under `sphinx.util.logging.logging`), and `sphinx.util.logging.LEVEL_NAMES` clearly map those of Python's `logging`. Those are still implementation details, yet further down we get this:

```python
    logger = logging.getLogger(__name__)
             ^^^^^^^
    [...]
    if self.__parser_verbosity(error.level) <= self.env.app.verbosity:
        self.logger.log(error.level, error.get_message(), location=(self.env.docname, self.lineno))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So, if I followed the code correctly, we're using the `sphinx.util` logger (naturally), but specifying the log level with Python's `logging` constants. At the very least, I'd use `sphinx.util.logging.logging.ERROR` and so on, but even that is inferior to using `sphinx.util.logging.LEVEL_NAMES` in my view as nothing prevents Sphinx from registering its own log levels.

BrunoMSantos (changes requested):
Nice set of changes and overall clean-up! I have only that one comment, which I believe is actionable.

hawkmoth/__init__.py
jnikula: > As far as I can see, you don't save nearly enough verbosity to justify the obfuscation of renaming one of the core modules. So I'd rather do `from sphinx.util import logging as log` all other considerations aside.

This was just to avoid a name collision with the Sphinx package, not to make it shorter, really. I only needed it for `log.ERROR` and `log.WARNING`. But I agree it's less than stellar.

jnikula: With the rest, the part I struggle with is figuring out what is Sphinx API and what is implementation details. Are the log level values API? Are the log level names API? Is the `sphinx.util.logging.LEVEL_NAMES` dictionary API?

The API documentation https://www.sphinx-doc.org/en/master/extdev/logging.html doesn't really say anything about any of them.

The `SphinxLoggerAdapter` implementation does not override the `LoggerAdapter` `info()`, `warning()`, etc. methods, and based on the Python documentation https://docs.python.org/3/library/logging.html#loggeradapter-objects they use the corresponding Python `logging` log levels. Sure, Sphinx could override them and use different levels, but to me it seems far fetched.

BrunoMSantos: > This was just to avoid a name collision with the Sphinx package [...]

I got that. Just bear in mind you can alias the other one instead and keep the standard one with its original name, which might be preferable.

> [...] Sure, Sphinx could override them and use different levels,  but to me it seems far fetched.

Yeah, I would also bet this will never be a problem, and yet it feels kind of dirty to do it this way. Maybe doing `logging.logging.ERROR` is an acceptable compromise? Or maybe make two imports like so:

```python
from sphinx.util import logging as log
from sphinx.util.logging import logging
```

It would at least remove the blind assumption that Sphinx logger is based on the standard Python logger.

jnikula: I wanted to remove the `ErrorLevel` enum because I thought it was redundant, and forced us to map Clang diagnostics to `ErrorLevel` and `ErrorLevel` to Sphinx error levels, i.e. two maps.

Since I don't want the parser to know anything about Sphinx, nor do I want the extension to know (much) about Clang, I think it stands to reason to use Python logging error levels in the parser. But based on what you're saying, we still need to map those to Sphinx error levels in the extension. Even though for all practical purposes they match.

I think I'll need to give this some more thought.

jnikula: Added new pull request #98 which I hope addresses all the concerns here. Fingers crossed. ;)

jnikula: As always, the review comments are much appreciated!

PR #98, opened 2022-12-05
Diagnostics v2

jnikula:
This superseeds #97.

I've switched the order of commits 2 and 3, and addressed the review concerns.

Additionally, new commit 4 drops the verbosity filtering/mapping, and lets Sphinx do that part. This makes parts of commit 3 temporary in the interest of not making functional changes in that.

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -8,6 +8,7 @@ Sphinx C Domain autodoc directive extension.
 """
 
 import glob
+import logging
 import os
 
 from docutils import nodes, statemachine
@@ -15,9 +16,9 @@ from docutils.parsers.rst import directives
 from docutils.statemachine import ViewList
 from sphinx.util.nodes import nested_parse_with_titles
 from sphinx.util.docutils import switch_source_input, SphinxDirective
-from sphinx.util import logging
+from sphinx.util import logging as sphinx_logging
 
-from hawkmoth.parser import parse, ErrorLevel
+from hawkmoth.parser import parse
 from hawkmoth.util import doccompat, strutil
 from hawkmoth import docstring
 
@@ -26,7 +27,7 @@ with open(os.path.join(os.path.abspath(os.path.dirname(__file__)),
     __version__ = version_file.read().strip()
 
 class CAutoBaseDirective(SphinxDirective):
-    logger = logging.getLogger(__name__)
+    logger = sphinx_logging.getLogger(__name__)
 
     option_spec = {
         'transform': directives.unchanged_required,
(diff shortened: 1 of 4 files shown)


BrunoMSantos (changes requested):
I'm afraid you haven't addressed my complaint at all :sweat_smile: 

Note this commit simply adds a round trip between numerical log levels and level names, all according to the standalone `logging` import, which an onlooker may _still_ question how it relates to the Sphinx logger. But not all is lost! This version makes mappings a lot clearer than in the original in my opinion. It's only a small change away from being perfect to my eyes.

I would suggest that we make `error.level` a string since we end up mapping these levels to the strings `'WARNING'`, `'ERROR'`, etc. anyway, and avoid importing `logging` altogether. Am I missing any reason to keep `error.level` an integer?

Side note, by using strings we can directly map to Sphinx specific levels that do not otherwise exist in `logging` like `'SEVERE'` and `'VERBOSE'`. Kind of like you do already in `__display_parser_diagnostics()` in the following commit and avoid the mapping table you have there.

See a rough sketch of what it would look like here: [diagnostics-v2-suggestion](https://github.com/BrunoMSantos/hawkmoth/commit/385178a555dfcf5a9cdf37bbcc4401b9bbd8f2b4)

hawkmoth/__init__.py
```diff
@@ -39,21 +39,18 @@ class CAutoBaseDirective(SphinxDirective):
...
-        # Map parser diagnostic level to Sphinx verbosity and level name
+        # Map parser diagnostic level to Sphinx level name
         log_level_map = {
-            logging.DEBUG: (3, 'DEBUG'),
-            logging.INFO: (3, 'INFO'),
-            logging.WARNING: (1, 'WARNING'),
-            logging.ERROR: (0, 'ERROR'),
-            logging.CRITICAL: (0, 'CRITICAL'),
+            logging.DEBUG: 'DEBUG',
+            logging.INFO: 'VERBOSE',
```
BrunoMSantos, line 45: I dropped this change in my suggestion by mistake only, but you get the idea. Feel free to add it back, it makes sense to me.

jnikula:
> I'm afraid you haven't addressed my complaint at all

Then I must say I haven't understood your complaint correctly!

> Note this commit simply adds a round trip between numerical log levels and level names, all according to the standalone `logging` import, which an onlooker may _still_ question how it relates to the Sphinx logger. But not all is lost! This version makes mappings a lot clearer than in the original in my opinion. It's only a small change away from being perfect to my eyes.
> 
> I would suggest that we make `error.level` a string since we end up mapping these levels to the strings `'WARNING'`, `'ERROR'`, etc. anyway, and avoid importing `logging` altogether. Am I missing any reason to keep `error.level` an integer?
> 
> Side note, by using strings we can directly map to Sphinx specific levels that do not otherwise exist in `logging` like `'SEVERE'` and `'VERBOSE'`. Kind of like you do already in `__display_parser_diagnostics()` in the following commit and avoid the mapping table you have there.
> 
> See a rough sketch of what it would look like here: [diagnostics-v2-suggestion](https://github.com/BrunoMSantos/hawkmoth/commit/385178a555dfcf5a9cdf37bbcc4401b9bbd8f2b4)

The problem I have with this is that it very subtly depends on using the Sphinx level names in parser code, which is supposed to be ignorant about Sphinx.

If re-using Python logging levels for ParserError feels wrong, we might just as well keep the ErrorLevel enum for it. My thinking was that it's better to use what's already a defined part of Python API instead of rolling our own.

jnikula:
> Hmm, thanks for the detail. Actually, I may have misunderstood what you were trying to achieve earlier.

Communication usually fails, except by accident. :)

> On the other hand, I now understanding you weren't overlooking anything and v2 really is the best version of what you intended... Can't say I'm sold, but I'm willing to concede if you still prefer it this way.

Let me put it this way: If I kept everything else the same, but the choice was between using Python logging error codes and preserving the current `ErrorLevel` enum (augmented to have all the levels), which one would you prefer?

BrunoMSantos:
> > On the other hand, I now understanding you weren't overlooking anything and v2 really is the best version of what you intended... Can't say I'm sold, but I'm willing to concede if you still prefer it this way.
> 
> Let me put it this way: If I kept everything else the same, but the choice was between using Python logging error codes and preserving the current `ErrorLevel` enum (augmented to have all the levels), which one would you prefer?

Oh, enum, definitely. We're not doing anything with the `logging` module that justifies its inclusion in my opinion.

jnikula:
> Oh, enum, definitely. We're not doing anything with the `logging` module that justifies its inclusion in my opinion.

I'll re-spin, and let's see if we've found a suitable compromise! :)

Issue #68, opened 2021-09-30
Clean up examples section in documentation


jnikula:
With the introduction of the more fine-grained `c:auto*` directives, the `doc/examples.rst` page has become a mess, and is in need of cleanup. It's not as helpful as it could be.

One idea is to have a single, sensible source file, displayed only once, maybe collapsible at the top, followed by only directive/output pairs.

jnikula:
There's always room for improvement, but it's better now, closing.

PR #102, opened 2022-12-11
Examples update v2

jnikula:
v2 of #101

I had more ambitious plans, but ran out of steam a bit. Regardless, I think this is a significant improvement in the examples section in the documentation.

All the changes here are under `test/` and the auto-generated `doc/examples.rst`. It's not really functionally significant at all.

@BrunoMSantos I'm not sure if it's really worth reviewing carefully, and I'm not sure I'm up for much more refactoring on this, unless I've made some bad blunders. If you are up for reviewing at all, I think the main thing would be running `make html` before and after, and checking the generated examples page. Again, I think there's more to be done, but I'm leaving it for future, and I think this is an improvement.

And if you're not up for reviewing this, I also don't think it's a big deal. I think I'm just going to merge and do a release.


jnikula:
Also, fixes #68.

BrunoMSantos (approved):
Alright, I agree, the output is better. I didn't fuss about individual commits, but it all makes sense to me still.

jnikula:
Much appreciated, @BrunoMSantos! Good luck with the moving!

Issue #52, opened 2021-06-17
cautodoc_transformations does not "pickle" nicely


jnikula:
Sphinx automatically pickles and caches the configuration, speeding up processing if there are no changes. Unfortunately, pickled function references in `cautodoc_transformations` are bound to change. This leads to re-reading of the source documents:

```
updating environment: [config changed ('cautodoc_transformations')] 5 added, 0 changed, 0 removed
reading sources... [100%] syntax
```

This is quite an unfortunate unintentional consequence of the transformation mechanism.

jnikula:
One idea is to use Sphinx events for the transformations, so that you could actually write independent Sphinx extensions for the transformations. `app.add_event('cautodoc-transform')` and `app.emit_firstresult('cautodoc-transform', comment, transform)` in Hawkmoth side, and `app.connect('cautodoc-transform', transformation_function)` in the extension implementing the transformation.

It's a bit sad I didn't think of this before implementing the current transformations, leading to another round of deprecations and changes.

PR #109, opened 2023-01-04
Random clean ups

BrunoMSantos:
Collection of changes I stumbled on while working on C++ support that are not (strictly) related to C++ and thus can be merged ahead.

I think the only contentious / tricky one is the parser fix 241cda39432a67529d72f2be6edc2fa63523449f. In the past I remember wanting the original behaviour somehow, but as I explain in that commit, I don't think we should do it, or at least not without a much larger effort to make sure we have consistent behaviour across the board.

Note this commit, if accepted, will replace the commit 49827c2ae5c0e4258e20beab1635e46290156020 in the C++ branch entirely.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -60,7 +60,7 @@ try:
         config = Config(napoleon_use_rtype=False)
         return str(docstring.GoogleDocstring(comment, config))
 
-    cautodoc_transformations = {
+    hawkmoth_transformations = {
         'napoleon': napoleon_transform,
         'javadoc-liberal': doccompat.javadoc_liberal,
     }
@@ -71,7 +71,7 @@ except ImportError:
     # into the documentation instead of generating.
     extensions.append('automock')
 
-cautodoc_root = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../test')
+hawkmoth_root = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../test')
 
 # Add any paths that contain templates here, relative to this directory.
 # templates_path = ['_templates']
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -25,7 +25,7 @@ with open(os.path.join(os.path.abspath(os.path.dirname(__file__)),
                        'VERSION')) as version_file:
     __version__ = version_file.read().strip()
 
-class CAutoBaseDirective(SphinxDirective):
+class _AutoBaseDirective(SphinxDirective):
     logger = logging.getLogger(__name__)
 
     option_spec = {
@@ -153,7 +153,7 @@ class CAutoBaseDirective(SphinxDirective):
 
         return node.children
 
-class CAutoDocDirective(CAutoBaseDirective):
+class _AutoDocDirective(_AutoBaseDirective):
     """Extract all documentation comments from the specified files"""
 
     # Allow passing a variable number of file patterns as arguments
@@ -176,13 +176,13 @@ class CAutoDocDirective(CAutoBaseDirective):
                                         location=(self.env.docname, self.lineno))
 
 # Base class for named stuff
-class CAutoSymbolDirective(CAutoBaseDirective):
+class _AutoSymbolDirective(_AutoBaseDirective):
     """Extract specified documentation comments from the specified file"""
 
     required_arguments = 1
     optional_arguments = 0
 
-    option_spec = CAutoBaseDirective.option_spec.copy()
+    option_spec = _AutoBaseDirective.option_spec.copy()
     option_spec.update({
         'file': directives.unchanged_required,
     })
@@ -201,41 +201,44 @@ class CAutoSymbolDirective(CAutoBaseDirective):
     def _get_names(self):
         return [self.arguments[0]]
 
-class CAutoVarDirective(CAutoSymbolDirective):
-    _docstring_types = [docstring.VarDocstring]
-
-class CAutoTypeDirective(CAutoSymbolDirective):
-    _docstring_types = [docstring.TypeDocstring]
-
-class CAutoMacroDirective(CAutoSymbolDirective):
-    _docstring_types = [docstring.MacroDocstring, docstring.MacroFunctionDocstring]
-
-class CAutoFunctionDirective(CAutoSymbolDirective):
-    _docstring_types = [docstring.FunctionDocstring]
-
-def members_filter(argument):
+def _members_filter(argument):
     # Use None for members option without an argument to not filter.
     if argument is None:
         return None
     return strutil.string_list(argument)
 
-class CAutoCompoundDirective(CAutoSymbolDirective):
-    option_spec = CAutoSymbolDirective.option_spec.copy()
+class _AutoCompoundDirective(_AutoSymbolDirective):
+    option_spec = _AutoSymbolDirective.option_spec.copy()
     option_spec.update({
-        'members': members_filter,
+        'members': _members_filter,
     })
 
     def _get_members(self):
         # By default use [] as a filter that does not match any members.
         return self.options.get('members', [])
 
-class CAutoStructDirective(CAutoCompoundDirective):
+class CAutoDocDirective(_AutoDocDirective):
+    pass
+
+class CAutoVarDirective(_AutoSymbolDirective):
+    _docstring_types = [docstring.VarDocstring]
+
+class CAutoTypeDirective(_AutoSymbolDirective):
+    _docstring_types = [docstring.TypeDocstring]
+
+class CAutoMacroDirective(_AutoSymbolDirective):
+    _docstring_types = [docstring.MacroDocstring, docstring.MacroFunctionDocstring]
+
+class CAutoFunctionDirective(_AutoSymbolDirective):
+    _docstring_types = [docstring.FunctionDocstring]
+
+class CAutoStructDirective(_AutoCompoundDirective):
     _docstring_types = [docstring.StructDocstring]
 
-class CAutoUnionDirective(CAutoCompoundDirective):
+class CAutoUnionDirective(_AutoCompoundDirective):
     _docstring_types = [docstring.UnionDocstring]
 
-class CAutoEnumDirective(CAutoCompoundDirective):
+class CAutoEnumDirective(_AutoCompoundDirective):
     _docstring_types = [docstring.EnumDocstring]
 
 def setup(app):
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -37,6 +37,25 @@ class _AutoBaseDirective(SphinxDirective):
 
     _docstring_types = None
 
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        # Map deprecated options; to be removed.
+        if self.env.config.cautodoc_root:
+            self.env.config.hawkmoth_root = self.env.config.cautodoc_root
+            fmt = 'cautodoc_root option is deprecated, please use hawkmoth_root.'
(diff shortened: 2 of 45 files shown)


BrunoMSantos:
> I think you've missed some of the more detailed comments on these changes that I already gave in #108. How do you want to proceed?

Hmm, I don't see any comments apart from the top ones, I thought you hadn't made any. Can you link one of them?

jnikula:
> Hmm, I don't see any comments apart from the top ones, I thought you hadn't made any. Can you link one of them?

Argh. They were in "pending" state. My bad. Posted them now. I hope.

BrunoMSantos:
> > Hmm, I don't see any comments apart from the top ones, I thought you hadn't made any. Can you link one of them?
> 
> Argh. They were in "pending" state. My bad. Posted them now. I hope.

I see them now, thanks! I'll have a look.

BrunoMSantos:
FYI, I think I either made the suggested changes or provided some counter point to all your comments and pushed new stuff.

The changes to this branch were minimal (a couple of typos, one of them in a commit message). I did push an additional commit at the end for better error handling in the command line tool.

I also pushed a small demo for class documentation, though that doesn't really affect this particular PR.

I keep thinking PRs make the changes more apparent and tracks versions, hence the force push, but that doesn't seem to be the case... Hope it's not too bad. -v2 next time.

BrunoMSantos:
Aaaand, pushed 2 more clean-ups while at it.

hawkmoth/parser.py
```diff
@@ -88,25 +88,21 @@ def _comment_extract(tu):
...
-        cursor = token_cursor
-
-        # Note: current_comment may be None
-        if current_comment is not None and docstring.Docstring.is_doc(current_comment.spelling):
-            comments[cursor.hash] = current_comment
+        # Otherwise the current comment documents _this_ cursor. I.e.: not a top
+        # level comment.
+        if current_comment and docstring.Docstring.is_doc(current_comment.spelling):
+            comments[token_cursor.hash] = current_comment
         current_comment = None
```
jnikula, line 102: It bugs me that I can't figure out why I've added the `cursor` variable in the first place! I've got a branch with *all* the history dating back to when I was just toying with using libclang from C, but there's no explanation. I give up. :)

hawkmoth/__init__.py
```diff
@@ -37,6 +37,25 @@ class _AutoBaseDirective(SphinxDirective):
...
+        if self.env.config.cautodoc_root:
+            self.env.config.hawkmoth_root = self.env.config.cautodoc_root
+            fmt = 'cautodoc_root option is deprecated, please use hawkmoth_root.'
+            self.logger.warning(fmt, location=(self.env.docname, self.lineno))
+
+        if self.env.config.cautodoc_transformations:
+            fmt = 'cautodoc_transformations option is deprecated, please use hawkmoth_transformations.'  # noqa: E501
+            self.logger.warning(fmt, location=(self.env.docname, self.lineno))
+            self.env.config.hawkmoth_transformations = self.env.config.cautodoc_transformations
+
```
jnikula, line 53: The thing that bugs me about this one is that I think I screwed up the whole cautodoc_transformations thing, and it needs to be changed, again. (Read: It bugs me because it rubs my own mistakes in my face.)

The problem is that the configuration is pickled for subsequent Sphinx runs, and function references don't pickle nicely.

I think the solution is to turn transformations to Sphinx events, which does have the nice property of being able to extend Hawkmoth with proper Sphinx extensions that listen to those events.

If we can squeese that into the next release, along with this and the C++ stuff, there aren't going to be so many deprecation and rename steps. If not, it's a nasty experience for the users I'm afraid.

hawkmoth/__init__.py
```diff
@@ -37,6 +37,25 @@ class _AutoBaseDirective(SphinxDirective):
...
     _docstring_types = None
 
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        # Map deprecated options; to be removed.
+        if self.env.config.cautodoc_root:
+            self.env.config.hawkmoth_root = self.env.config.cautodoc_root
+            fmt = 'cautodoc_root option is deprecated, please use hawkmoth_root.'
+            self.logger.warning(fmt, location=(self.env.docname, self.lineno))
```
jnikula, line 47: What does `location=(self.env.docname, self.lineno)` reference and does it make sense for this?

doc/conf.py
```diff
@@ -71,7 +71,7 @@ def napoleon_transform(comment):
     # into the documentation instead of generating.
     extensions.append('automock')
 
-cautodoc_root = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../test')
+hawkmoth_root = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../test')
```
jnikula, line 75: Please also fix `test/conf.py`.

hawkmoth/__init__.py
jnikula: So it does reference the where the extension is used, but it sure emits a **lot** of warnings.

hawkmoth/docstring.py
```diff
@@ -188,15 +188,15 @@ class TypeDocstring(Docstring):
```
jnikula, line 188: I like the simplicity of the code and the predictability and requirement to be explicit.

I do need to come with terms about documenting typedeffed anonymous structs, which I think is the main use case here. This change basically requires to you document a named struct and the typedef separately. And it requires you to name the struct when it's otherwise unnecessary. And no, the current thing as exemplified in the commit message is not great either.

I also hope it doesn't break a lot of existing documentation people might have.

hawkmoth/__init__.py
BrunoMSantos: Hmm, I haven't really thought about that. I remember liking the changes back then and I may not know enough of Python to really grasp the issues you're referring to.

Do you have an idea of how it should / could be done instead?

hawkmoth/docstring.py
BrunoMSantos: I really wanted us to allow that use case too. Off the top of my head, we could do something like:

Does it work?

```rst
.. c:type:: some_typedef

   .. c:member:: member
```

Maybe this?

```rst
.. c:type:: some_typedef
   
   .. c:struct:: @anonymous_<hash>

      .. c:member:: member
```

Or this one?

```rst
.. c:type:: some_typedef
   
   :c:struct:`anonymous_<hash>`

.. c:struct:: @anonymous_<hash>

   .. c:member:: member
```

Even then there's still the inconsistency with instantiation of multiple objects / variables of a given anonymous type, and that has no good solution I think.

Either way, these require much more extensive rework and this commit at least removes some odd quirks. A net improvement I'd say. That said, we can also hold this one back for now and let it sink a bit longer if you're not ready for it :sweat_smile:

hawkmoth/__init__.py
BrunoMSantos: I confess I copied this pattern here. Maybe the other deprecations were less noisy because they were not ubiquitous. E.g.: we don't use transformations everywhere, but we use the root directory for _every_ _single_ _directive_.

Maybe there's some hook in Sphinx in which we could validate `conf.py` variables only once?

hawkmoth/__init__.py
jnikula: I liked the changes way back too, but only later realized every no-op rebuild causes:

```
updating environment: [config changed ('hawkmoth_transformations')] 7 added, 0 changed, 0 removed
```

which causes the rewrite of the output, even when nothing changed.

See issue #52.

hawkmoth/__init__.py
jnikula: Looks like the default parameter of `add_config_value()` can be callable: https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_config_value

This works, but might also be an incredibly stupid thing to do:

```
def _deprecate(conf, old, new, default=None):
    if conf[old]:
        logger = logging.getLogger(__name__)
        logger.warning(f'{old} is deprecated in favour of {new}')
        conf[new] = conf[old]
        del conf[old]
        return conf[new]

    return default

def setup(app):
    ...
    app.add_config_value('hawkmoth_root',
                         lambda conf: _deprecate(conf, 'cautodoc_root', 'hawkmoth_root', app.confdir),
                         'env', [str])
    ...

```

hawkmoth/__init__.py
BrunoMSantos: Ah, I get it now. Good point...

So, it's unfortunate if we deprecate this now only to deprecate it again soon when we address that issue. It's also a bit conspicuous to rename every variable except this one. And fixing that issue 1st would just stretch me a bit too thin for what I had in mind.

It may be an easy fix, but honestly I have no idea. Is there an interim solution you'd accept here?

jnikula:
@BrunoMSantos Process wise, how do you feel about me picking up individual commits that I think are good to go, and just pushing them, to avoid several rounds with the pull requests?

hawkmoth/__init__.py
jnikula: I've got some ideas and code for the fix myself, but it might take a while before I get around to polishing it. I guess the question is, do you have a deadline for having a release with the features you need? I don't mind having multiple deprecation steps *between* releases, but I'd really like to get this sorted out before the next release.

Another interim solution might be just not touching this for now, even if the name is inconsistent.

BrunoMSantos:
> @BrunoMSantos Process wise, how do you feel about me picking up individual commits that I think are good to go, and just pushing them, to avoid several rounds with the pull requests?

I'm all for it! Just make sure to `make test` at every stage just in case there's no order or inter dependency.

jnikula:
> I'm all for it! Just make sure to `make test` at every stage just in case there's an order or inter dependency.

Pushed a handful. Dropped a couple of stale comments while at it. One commit conflicted, but it was a trivial fix. It'll hit you when you rebase.

hawkmoth/__init__.py
BrunoMSantos: > I've got some ideas and code for the fix myself, but it might take a while before I get around to polishing it.

Nice!

> I guess the question is, do you have a deadline for having a release with the features you need? I don't mind having multiple deprecation steps _between_ releases, but I'd really like to get this sorted out before the next release.

Well, it's not a given that we'll take this route. Bottom line, I'd like to have something within a week that competes with breathe, which I'm still optimistic about. Other alternatives would imply moving Sphinx documentation to something else entirely, so I'm placing lower bets on those.

Technically, I think there would be no competition if not for the fear of taking a dependency on an untested solution.

And of course getting the changes upstreamed and an official release in pypi would be important in building that confidence. Not trying to put any pressure on you as my reviewer! :sweat_smile: Just bearing out the facts.

> Another interim solution might be just not touching this for now, even if the name is inconsistent.

I was actually thinking the same. Thought I'd prefer to use the new variables on new projects if we already know we want them. I can go either way honestly...

BrunoMSantos: Unrelated note: I forgot to update the docs with the new variables... So this is incomplete anyway.

hawkmoth/__init__.py
BrunoMSantos: Thanks for the snippet! I'll have a look.

BrunoMSantos: Going by the docs, I'd say this is pretty neat and an acceptable way of doing this. I'll take this approach in v2.

Only issue I see is that not all deprecations are simple renames, which would complicate this a bit if it mattered in our case. In the limit we'd need one 'deprecation' function for each variable, which I think is ok. We can cross that bridge when we get there of course.

BrunoMSantos:
Closing. I'll create a new PR with the remaining commits so I don't mess with the comments here.

PR #110, opened 2023-01-07
Compat removal and splitlines cleanup RFC

jnikula:
This is prep work for #52. The sensible thing for any transformation extension via events is passing references to lists, which the extension modifies in-place, instead of passing and returning multi-line strings. So we start modifying docstring to also operate on lists of strings.

@BrunoMSantos This is not intended to have priority over what you're doing, but I'd like you to check if the changes in docstring are compatible with what you're doing. I can rebase this later.


jnikula:
In particular, "docstring: reduce the number of line splitting and joining" is the one I'm seeking feedback on. It drops `{text}` from `_fmt` and starts handling the text and the header (or Sphinx C domain directive) separately.

BrunoMSantos:
I like the changes and I don't think it affects me in any way: simple rebases and nothing really fundamental changes. I'd be ok with this going ahead right away minus a few nitpicks.

test/test_cli.py
```diff
@@ -49,13 +49,9 @@ def _get_output(testcase, monkeypatch, capsys, **options):
 
     options = options.get('directive-options', {})
 
-    transform = options.get('compat', None)
```
BrunoMSantos, line 52: Maybe do this separately and earlier in a `test: stop using deprecated stuff`?

hawkmoth/__init__.py
```diff
@@ -10,7 +10,7 @@
 import glob
```
BrunoMSantos, line 10: Penultimate sentence in the commit looks a bit funny: as ..., so ...
Otherwise this looks like a good set of changes too.

hawkmoth/docstring.py
```diff
@@ -99,10 +99,11 @@ def is_doc(comment):
...
         return comment.startswith('/**') and comment != '/**/'
 
-    def _get_plain_comment(self):
-        """Return plain comment with comment markers and line prefixes removed."""
-        lines = statemachine.string2lines(self._text, 8, convert_whitespace=True)
+    def _get_comment_lines(self):
+        return statemachine.string2lines(self._text, 8, convert_whitespace=True)
 
+    def _remove_comment_markers(self, lines):
+        """Remove comment markers and line prefixes from comment lines."""
```
BrunoMSantos, line 106: This one can be a static method too. Not that we'll be using it as such I guess.

hawkmoth/docstring.py
```diff
@@ -164,6 +149,24 @@ def get_docstring(self, transform=None):
                                   type_spacer=spacer, args=args)
         lines[:0] = header.splitlines()
```
BrunoMSantos, line 150: I thought this slice was a bit obfuscating before, but now it's pretty bad. Basically we imply here that `_insert_header_lines` is called _after_ we've indented the remaining lines and _before_ we nest them. Both of which occur outside of this method.

Maybe do a `_get_header_lines` instead? That way the final line stitching is done all in the same place.

jnikula:
> I like the changes and I don't think it affects me in any way: simple rebases and nothing really fundamental changes. I'd be ok with this going ahead right away minus a few nitpicks.

Thanks! I'll fix them. I've also got some follow-ups on top. We'll soon be having more design decisions to screw up with the parameters to send to the event handlers! ;D

hawkmoth/docstring.py
jnikula: Wasn't too happy with it myself.

Any idea if there's a perf impact for doing this:

```
lines = header.splitlines() + lines
```

instead of:

```
lines[:0] = header.splitlines()
```

?

BrunoMSantos: [Didn't test it myself, but seems there isn't any.](https://stackoverflow.com/questions/12088089/python-list-concatenation-efficiency).

This would be my rough take on `get_docstring`:

```python
        header = self._get_header_lines()
        text = self._get_comment_lines()

        # This could be inside `_get_comment_lines`
        self._remove_comment_markers(text)

        if transform is not None:
            ...

        self._nest(text, self._indent)
        lines = header + text
        self._nest(lines, self._nest)

        ...
        return lines
```

jnikula: Any preference on calling the static methods via self or Docstring?

BrunoMSantos: Hmm, I wrote `self` mostly out of habit. Only reason I can think of is that if it ever stops being static, `self` will still work and only external users would need fixing.

jnikula: Uh, I hit some issue with `self`, and just left the `Docstring` in for now.

jnikula:
v2 at #111 hopefully with all issues fixed, and an additional commit on top. This should let me proceed with transformations without any further impact on docstring.py, avoiding conflicts.

I don't know if it's better to force-push the same branch or push a new branch, opted for the latter.

PR #111, opened 2023-01-07
Compat removal and splitlines cleanup v2

jnikula:
v2 of #110


hawkmoth/docstring.py
```diff
@@ -130,19 +130,22 @@ def _nest(text, nest):
         Returns:
             str: Indented reST documentation string.
         """
-        return re.sub('(?m)^(?!$)', '   ' * nest, text)
+        lines[:] = [re.sub('^(?!$)', '   ' * nest, line) for line in lines]
```
BrunoMSantos: Sorry I missed it before, but the docstring of the function mentions a return value.

jnikula:
Btw I also considered removing the newlines from all the `_fmt` strings, and making them just the directive. Then `_get_header_lines()` would add the newlines. But I left that up to you and/or for the future.

BrunoMSantos:
> Btw I also considered removing the newlines from all the `_fmt` strings, and making them just the directive. Then `_get_header_lines()` would add the newlines. But I left that up to you and/or for the future.

That may be cleaner overall actually. It gets rid of the `splitlines()` for one thing.

It's nitpicking though, so I'll leave it up to you. This is a good set of changes regardless. +1

hawkmoth/docstring.py
jnikula: Force pushed with the fix.

For a code documentation tool we have pretty lousy code documentation ourselves. :D

PR #112, opened 2023-01-07
Transform via Sphinx/Hawkmoth extensions

jnikula:
FYI @BrunoMSantos, this is what I'm drafting. Please don't waste review cycles on this yet, it's just to give you an idea.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -48,22 +48,15 @@ with open(os.path.join(os.path.abspath(os.path.dirname(__file__)),
 # This is not a good example to follow in regular documentation.
 try:
     import hawkmoth
-    from hawkmoth.util import doccompat, readthedocs
-    from sphinx.ext.napoleon import docstring, Config
+    from hawkmoth.util import readthedocs
 
     readthedocs.clang_setup()
 
     extensions.append('hawkmoth')
+    extensions.append('hawkmoth.ext.napoleon')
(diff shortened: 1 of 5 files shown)


BrunoMSantos:
I see. It fits in! And you're planning this for the next release, is that right? Wondering if I should just leave `cautodoc_transformations` alone in the other branch.

jnikula:
> I see. It fits in! And you're planning this for the next release, is that right? Wondering if I should just leave `cautodoc_transformations` alone in the other branch.

I guess more than anything I'd like to avoid first renaming `cautodoc_transformations` to `hawkmoth_transformations`, with deprecations, and then deprecating the both of them in favour of this. There's been enough confusion around this feature already.

The primary uncertainty I have around this is the signature of the event. What should be passed to `hawkmoth-process-docstring`? I admit my Python knowledge falls short of how to ensure forward/backward compatibility of an interface like this.

jnikula:
I realized it's possible to implement the `cautodoc_transformations` stuff as a built-in extension using the `hawkmoth-process-docstring` event. I think it's pretty clean this way (pushed the implementation). I think `cautodoc_transformations` should stay, and only the other configs should be converted to `hawkmoth_*`.

BrunoMSantos:
Sounds good. I'll rebase my work accordingly.

PR #113, opened 2023-01-09
Random clean-ups v2

BrunoMSantos:
Re-spin on the commits from #109 that did not go through.

It comprises 3 changes:

1. `parser: fix anonymous type detection` is the most contentious one. I still think what we have is 'broken' anyway so might as well be annoying but consistent / accurate for the time being.
This is the only one that has a bigger influence in the `cpp-initial-support` series, but I have an alternative fix if we opt for the status quo. That fix would belong to the other series though as is C++ specific.

2. `hawkmoth: add name appropriate configuration options` and its related clean-ups assume that the transformation rework is not done yet and leave `cautodoc_transformations` alone. If that were to go 1st, there will likely be some conflicts to address and commit messages to review.
Otherwise I think we're aligned on these.

3. Finally, the indentation thing, it's a matter of consistency for me, but it's purely cosmetic. Feel free to drop it.

### DIFF
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -51,7 +51,7 @@ class _AutoBaseDirective(SphinxDirective):
                             location=(self.env.docname, self.lineno))
 
     def __get_clang_args(self):
-        clang_args = self.env.config.cautodoc_clang.copy()
+        clang_args = self.env.config.hawkmoth_clang.copy()
 
         clang_args.extend(self.options.get('clang', []))
 
@@ -151,7 +151,7 @@ class _AutoDocDirective(_AutoBaseDirective):
 
     def _get_filenames(self):
         for pattern in self.arguments:
-            filenames = glob.glob(os.path.join(self.env.config.cautodoc_root, pattern))
+            filenames = glob.glob(os.path.join(self.env.config.hawkmoth_root, pattern))
             if len(filenames) == 0:
                 self.logger.warning(f'Pattern "{pattern}" does not match any files.',
                                     location=(self.env.docname, self.lineno))
@@ -185,7 +185,7 @@ class _AutoSymbolDirective(_AutoBaseDirective):
                                 location=(self.env.docname, self.lineno))
             return []
 
-        return [os.path.abspath(os.path.join(self.env.config.cautodoc_root, filename))]
+        return [os.path.abspath(os.path.join(self.env.config.hawkmoth_root, filename))]
 
     def _get_names(self):
         return [self.arguments[0]]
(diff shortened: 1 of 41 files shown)


hawkmoth/__init__.py
```diff
@@ -63,7 +63,7 @@ def __get_transform(self):
 
         if transformations is None:
             if tropt is not None:
-                self.logger.warning('transform specified without cautodoc_transformations config.',
+                self.logger.warning('transform specified without hawkmoth_transformations config.',
```
jnikula, line 66: Need to drop this hunk.

hawkmoth/__init__.py
```diff
@@ -230,11 +230,33 @@ class CAutoUnionDirective(_AutoCompoundDirective):
...
 
+def _deprecate(conf, old, new, default=None):
+    if conf[old]:
+        logger = logging.getLogger(__name__)
+        logger.warning(f'{old} is deprecated in favour of {new}')
+        conf[new] = conf[old]
+        del conf[old]
+        return conf[new]
+    return default
+
```
jnikula, line 241: Need to double check modification of the config does not break Sphinx caching the environment for subsequent runs.

BrunoMSantos: Hmm, I thought it gated the compilation on `conf.py` being changed, but it's a legitimate concern.

I didn't notice any difference with regards to skipping compilation steps or not when compiling our own docs. How do you usually check this?

BrunoMSantos:
By the way, (some of) this work and following C++ stuff has been / is being done on my employer/client's dime ([Critical Software](www.criticalsoftware.com)).

They waved any copyright claims for this, which is why I kept pushing things as usual. The only thing they said was that if the contributions go anywhere, they'd welcome a small acknowledgement. I was thinking we could mention them in the release notes when we release C++ support regardless of whether we end up using it or not, though they don't seem fussed about it either way. Ultimately it's your decision.

hawkmoth/__init__.py
jnikula: I think history has shown I usually don't. :p

jnikula:
> By the way, (some of) this work and following C++ stuff has been / is being done on my employer/client's dime ([Critical Software](www.criticalsoftware.com)).
> 
> They waved any copyright claims for this, which is why I kept pushing things as usual. The only thing they said was that if the contributions go anywhere, they'd welcome a small acknowledgement. I was thinking we could mention them in the release notes when we release C++ support regardless of whether we end up using it or not, though they don't seem fussed about it either way. Ultimately it's your decision.

I appreciate it, and that sounds fine.

hawkmoth/__init__.py
jnikula: Played with it a bit, and I think it's okay.

jnikula:
I'm slightly sad about the `typedef struct { ... } T;` part, but I think it's the way to go.

BrunoMSantos:
Thanks for the review ;)

PR #116, opened 2023-01-10
C++ initial support

BrunoMSantos:
Finally, a more polished attempt at getting minimal support for C++ in. This is a re-spin of the commits from #108 that were not yet part of the clean-up series.

Took a while to get here, but not because the changes are that massive, on the contrary: all changes are fairly minor as per the comments in the RFC and it still makes a mess in the test folder, which I may try to tackle tomorrow as I wait for further comments, but which could also be left for a 2nd pass.

The reason it took so long was actually that I was working on the [`cpp-class-wip` branch](https://github.com/BrunoMSantos/hawkmoth/compare/cpp-initial-support...BrunoMSantos:hawkmoth:cpp-class-wip) as a preview of what's to come for better context. There are still a few things that need polish, but I'm actually pretty excited with how far I took it already. Hopefully it will make a good case :)

Note that there are some issues in #108 that were never marked resolved. @jnikula, you may still want to follow up on some of those.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,18 +20,24 @@
 Hawkmoth - Sphinx Autodoc for C
 ===============================
 
-Hawkmoth is a minimalistic Sphinx_ `C Domain`_ autodoc directive extension to
-incorporate formatted C source code comments written in reStructuredText_ into
-Sphinx based documentation. It uses Clang Python Bindings for parsing, and
-generates C Domain directives for C API documentation, and more. In short,
-Hawkmoth is Sphinx Autodoc for C.
+Hawkmoth is a minimalistic Sphinx_ `C/C++ Domain`_ autodoc directive extension
+to incorporate formatted C/C++ source code comments written in reStructuredText_
+into Sphinx based documentation. It uses Clang Python Bindings for parsing, and
+generates C/C++ Domain directives for C API documentation, and more. In short,
+Hawkmoth is Sphinx Autodoc for C/C++.
 
-Hawkmoth aims to be a compelling alternative for documenting C projects using
-Sphinx, mainly through its simplicity of design, implementation and use.
+.. attention::
+
+   C++ support is incomplete and highly experimental. Currently it supports C
+   syntax only and is definitely not usable, though we welcome testers and
+   bug reports / feature requests.
+
+Hawkmoth aims to be a compelling alternative for documenting C/C++ projects
+using Sphinx, mainly through its simplicity of design, implementation and use.
 
 .. _Sphinx: http://www.sphinx-doc.org
 
-.. _C Domain: http://www.sphinx-doc.org/en/stable/domains.html
+.. _C/C++ Domain: http://www.sphinx-doc.org/en/stable/domains.html
 
 .. _reStructuredText: http://docutils.sourceforge.net/rst.html
 
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -34,6 +34,7 @@ class _AutoBaseDirective(SphinxDirective):
     }
     has_content = False
 
+    _domain = None
     _docstring_types = None
 
     def __display_parser_diagnostics(self, errors):
@@ -100,7 +101,8 @@ class _AutoBaseDirective(SphinxDirective):
         # Tell Sphinx about the dependency
         self.env.note_dependency(filename)
 
-        docstrings, errors = parse(filename, clang_args=clang_args)
+        docstrings, errors = parse(filename, domain=self._domain,
+                                   clang_args=clang_args)
 
         self.__display_parser_diagnostics(errors)
 
@@ -207,27 +209,34 @@ class _AutoCompoundDirective(_AutoSymbolDirective):
         return self.options.get('members', [])
 
 class CAutoDocDirective(_AutoDocDirective):
-    pass
+    _domain = 'c'
 
 class CAutoVarDirective(_AutoSymbolDirective):
+    _domain = 'c'
     _docstring_types = [docstring.VarDocstring]
 
 class CAutoTypeDirective(_AutoSymbolDirective):
+    _domain = 'c'
     _docstring_types = [docstring.TypeDocstring]
 
 class CAutoMacroDirective(_AutoSymbolDirective):
+    _domain = 'c'
     _docstring_types = [docstring.MacroDocstring, docstring.MacroFunctionDocstring]
 
 class CAutoFunctionDirective(_AutoSymbolDirective):
+    _domain = 'c'
     _docstring_types = [docstring.FunctionDocstring]
 
 class CAutoStructDirective(_AutoCompoundDirective):
+    _domain = 'c'
     _docstring_types = [docstring.StructDocstring]
 
 class CAutoUnionDirective(_AutoCompoundDirective):
+    _domain = 'c'
     _docstring_types = [docstring.UnionDocstring]
 
 class CAutoEnumDirective(_AutoCompoundDirective):
+    _domain = 'c'
     _docstring_types = [docstring.EnumDocstring]
 
 def _deprecate(conf, old, new, default=None):
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -87,7 +87,10 @@ class _AutoBaseDirective(SphinxDirective):
             lines[:] = [line for line in text.splitlines()]
 
     def __parse(self, filename):
-        clang_args = self.__get_clang_args()
+        # Always pass `-xc++` to the compiler on 'cpp' domain as the first
+        # option so that the user can override it.
+        clang_args = ['-xc++'] if self._domain == 'cpp' else []
+        clang_args.extend(self.__get_clang_args())
 
         # Cached parse results per rst document
         parsed_files = self.env.temp_data.setdefault('cautodoc_parsed_files', {})
--- a/hawkmoth/__init__.py
+++ b/hawkmoth/__init__.py
@@ -239,6 +239,37 @@ class CAutoEnumDirective(_AutoCompoundDirective):
     _domain = 'c'
     _docstring_types = [docstring.EnumDocstring]
 
+class CppAutoDocDirective(_AutoDocDirective):
+    _domain = 'cpp'
+
+class CppAutoVarDirective(_AutoSymbolDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.VarDocstring]
+
+class CppAutoTypeDirective(_AutoSymbolDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.TypeDocstring]
+
+class CppAutoMacroDirective(_AutoSymbolDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.MacroDocstring, docstring.MacroFunctionDocstring]
+
+class CppAutoFunctionDirective(_AutoSymbolDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.FunctionDocstring]
+
+class CppAutoStructDirective(_AutoCompoundDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.StructDocstring]
+
+class CppAutoUnionDirective(_AutoCompoundDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.UnionDocstring]
+
+class CppAutoEnumDirective(_AutoCompoundDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.EnumDocstring]
+
 def _deprecate(conf, old, new, default=None):
     if conf[old]:
         logger = logging.getLogger(__name__)
@@ -274,6 +305,15 @@ def setup(app):
     app.add_directive_to_domain('c', 'autoenum', CAutoEnumDirective)
     app.add_directive_to_domain('c', 'automacro', CAutoMacroDirective)
     app.add_directive_to_domain('c', 'autofunction', CAutoFunctionDirective)
+
+    app.add_directive_to_domain('cpp', 'autodoc', CppAutoDocDirective)
+    app.add_directive_to_domain('cpp', 'autovar', CppAutoVarDirective)
+    app.add_directive_to_domain('cpp', 'autotype', CppAutoTypeDirective)
+    app.add_directive_to_domain('cpp', 'autostruct', CppAutoStructDirective)
+    app.add_directive_to_domain('cpp', 'autounion', CppAutoUnionDirective)
+    app.add_directive_to_domain('cpp', 'autoenum', CppAutoEnumDirective)
+    app.add_directive_to_domain('cpp', 'automacro', CppAutoMacroDirective)
+    app.add_directive_to_domain('cpp', 'autofunction', CppAutoFunctionDirective)
 
     return dict(version=__version__,
                 parallel_read_safe=True, parallel_write_safe=True)
--- a/hawkmoth/__main__.py
+++ b/hawkmoth/__main__.py
@@ -30,7 +30,11 @@ def main():
     from FILE, along with the generated C Domain directives, to standard
     output. Include metadata with verbose output.""")
     parser.add_argument('file', metavar='FILE', type=filename, action='store',
-                        help='The C source or header file to parse.')
+                        help='The C/C++ source or header file to parse.')
+    parser.add_argument('--domain',
+                        choices=['c', 'cpp'],
+                        default='c',
+                        help='Sphynx domain to be used.')
     parser.add_argument('--compat',
                         choices=['none',
                                  'javadoc-basic',
@@ -44,6 +48,7 @@ def main():
     args = parser.parse_args()
 
     comments, errors = parse(args.file, clang_args=args.clang)
+    comments, errors = parse(args.file, domain=args.domain, clang_args=args.clang)
 
     for comment in comments.walk():
         if args.verbose:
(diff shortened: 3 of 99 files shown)


hawkmoth/parser.py
```diff
@@ -156,6 +156,14 @@ def _comment_extract(tu):
                                  CursorKind.MACRO_INSTANTIATION]:
             continue
 
+        # Cursors that are 1) never documented themselves, and 2) not allowed
+        # between the comment and the actual cursor being documented.
+        if token_cursor.kind in [CursorKind.UNEXPOSED_DECL]:
+            if current_comment and docstring.Docstring.is_doc(current_comment.spelling):
```
jnikula, line 162: Side note, `current_comment and docstring.Docstring.is_doc(current_comment.spelling)` could use a helper, it's already repeated quite a bit. Does *not* need to be part of this series.

hawkmoth/parser.py
```diff
@@ -413,7 +421,23 @@ def parse(filename, domain='c', clang_args=None):
...
+        # For some reason, the Python bindings don't return the cursor kind
+        # LINKAGE_SPEC as one would expect, so we need to do it the hard way.
+        if cursor.kind == CursorKind.UNEXPOSED_DECL:
+            tokens = cursor.get_tokens()
+            if next(tokens).spelling == 'extern':
+                nested_domain = 'c' if next(tokens).spelling == '"C"' else domain
+                for c in cursor.get_children():
+                    if c.hash in comments:
+                        result.add_children(_recursive_parse(nested_domain,
+                                                             comments, errors, c, 0))
```
jnikula, line 437: Does this not document the children of `UNEXPOSED_DECL` twice? For the first time as the children of `UNEXPOSED_DECL`, and the second time as the children of `tu.cursor.get_children()`?

jnikula: Also, would be nice to have a helper for identifying the `extern "<lang>"` part instead of inlining it at the highest level parser function. Can be fixed later too, though.

hawkmoth/parser.py
```diff
@@ -413,7 +421,23 @@ def parse(filename, domain='c', clang_args=None):
...
-        if cursor.hash in comments:
+
+        # Identify `extern "C"` and `extern "C++"` blocks and recursively parse
+        # their contents. Only `extern "C"` is of any relevance in choosing a
+        # different domain.
+        # For some reason, the Python bindings don't return the cursor kind
+        # LINKAGE_SPEC as one would expect, so we need to do it the hard way.
+        if cursor.kind == CursorKind.UNEXPOSED_DECL:
+            tokens = cursor.get_tokens()
+            if next(tokens).spelling == 'extern':
```
BrunoMSantos, line 432: Apparently there are some `UNEXPOSED_DECL` cursors that have 0 tokens. The 1st case I found was `__BEGIN_DECLS` and I didn't check for others. These showed up in some standard headers and I'm guessing they're implementation specific stuff which are meaningless syntax wise.

Anyway, I did not account for such ghost cursors here and this naturally crashes when `next` fails to return a valid token.

Regardless of ghost cursors, assuming any cursor would not only have at least 1 token, but 2 was a mistake, so a similar fix is needed for the following `next` as well.

I actually have a fix already, but I'm avoiding force pushes for now.

hawkmoth/__main__.py
```diff
@@ -30,7 +30,11 @@ def main():
     from FILE, along with the generated C Domain directives, to standard
     output. Include metadata with verbose output.""")
     parser.add_argument('file', metavar='FILE', type=filename, action='store',
-                        help='The C source or header file to parse.')
+                        help='The C/C++ source or header file to parse.')
+    parser.add_argument('--domain',
+                        choices=['c', 'cpp'],
+                        default='c',
+                        help='Sphynx domain to be used.')
```
jnikula, line 37: *Sphinx

hawkmoth/__main__.py
```diff
@@ -30,7 +30,11 @@ def main():
     from FILE, along with the generated C Domain directives, to standard
     output. Include metadata with verbose output.""")
     parser.add_argument('file', metavar='FILE', type=filename, action='store',
-                        help='The C source or header file to parse.')
+                        help='The C/C++ source or header file to parse.')
```
jnikula, line 33: Nitpick, I'm slightly averse to using "C/C++" anywhere, and would generally prefer "C or C++" or "C and C++", depending on the case.

hawkmoth/__init__.py
```diff
@@ -239,6 +239,37 @@ class CAutoEnumDirective(_AutoCompoundDirective):
...
+    _domain = 'cpp'
+    _docstring_types = [docstring.VarDocstring]
+
+class CppAutoTypeDirective(_AutoSymbolDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.TypeDocstring]
+
+class CppAutoMacroDirective(_AutoSymbolDirective):
+    _domain = 'cpp'
+    _docstring_types = [docstring.MacroDocstring, docstring.MacroFunctionDocstring]
```
jnikula, line 258: So the idea is we have `.. cpp:automacro:` but it ends up as `.. c:macro:` in reStructuredText?

test/test_cli.py
```diff
@@ -47,13 +47,17 @@ def _get_output(testcase, monkeypatch, capsys, **options):
     if directive:
         pytest.skip(f'{directive} directive test')
 
-    options = options.get('directive-options', {})
+    domain = options.get('domain', None)
```
jnikula, line 50: Isn't the `None` default superfluous?

test/testenv.py
```diff
@@ -25,6 +25,7 @@ def get_testcases(path):
             yield os.path.join(path, f)
 
 options_schema = strictyaml.Map({
+    'domain': strictyaml.Str(),
```
jnikula, line 28: Domain could be an Enum to limit the values: https://hitchdev.com/strictyaml/using/alpha/scalar/enum/

I guess it's cleaner to require adding the domain instead of making it optional with 'c' as default?

test/autoenum.yaml
```diff
@@ -1,3 +1,4 @@
+domain: c
```
jnikula, line 1: *Sphinx in the commit message

test/variable.c
```diff
@@ -41,7 +41,7 @@ const char* (*const volatile function_pointer_with_multiple_qualifiers)(const ch
 /**
  * a complex type involving function pointers somehow
  */
-const char* (*const *volatile *restrict legal_type_involving_function_pointer[12][3])(const char *in);
+const char* (*const *volatile *const legal_type_involving_function_pointer[12][3])(const char *in);
```
jnikula, line 44: 'restrict' is a case in `_type_fixup()` so I'd keep that, but if necessary split it to a separate test.

test/test_cli.py
BrunoMSantos: Hmm, it appears so. For some reason I have the habit of doing it. Maybe in older Python versions it would raise an exception? Anyway, I'll remove them, it's not the only one.

hawkmoth/__main__.py
BrunoMSantos: It's not even the 1st time :facepalm:

hawkmoth/__main__.py
BrunoMSantos: That's fair and I don't have any strong preference. I'll update it for a final round.

test/autoenum-cpp.yaml
```diff
@@ -0,0 +1,10 @@
+domain: cpp
+directive: autoenum
+directive-arguments:
+  - foo
+directive-options:
+  file: enum.c
+  members:
+    - bar
+  clang:
+    - -xc++
```
jnikula, line 10: Divided between having `-xc++` and adding similar heuristics as you added in the extension to pass that automatically for cpp domain.

README.rst
```diff
@@ -20,18 +20,24 @@
...
-Hawkmoth is a minimalistic Sphinx_ `C Domain`_ autodoc directive extension to
-incorporate formatted C source code comments written in reStructuredText_ into
-Sphinx based documentation. It uses Clang Python Bindings for parsing, and
-generates C Domain directives for C API documentation, and more. In short,
-Hawkmoth is Sphinx Autodoc for C.
+Hawkmoth is a minimalistic Sphinx_ `C/C++ Domain`_ autodoc directive extension
+to incorporate formatted C/C++ source code comments written in reStructuredText_
+into Sphinx based documentation. It uses Clang Python Bindings for parsing, and
+generates C/C++ Domain directives for C API documentation, and more. In short,
+Hawkmoth is Sphinx Autodoc for C/C++.
```
jnikula, line 27: I'm still frowning on the C/C++. Two languages, two domains... one wouldn't say "C/Python". :p

hawkmoth/__init__.py
BrunoMSantos: That was my idea, yes. Of course I would hope Sphinx eventually adds `cpp:macro`.

Even if that never happens, my arguments:
1. `cpp:autodoc` will produce `c:macro` anyway.
2. The user is going to bump into the inconsistency when referencing macros either way.
3. This avoids a special case for the domain detection, which is user facing API. (It also simplifies implementation.)

test/autoenum-cpp.yaml
BrunoMSantos: Hmm, so default to `-xc++` in `cpp` domain tests and override it with `-xc` for the wrong syntax test? Might be a good idea. Let me think about it, but I think there's no down side.

README.rst
BrunoMSantos: I didn't invent the C/C++ monogram xD

Though, sadly, C++ is only half a new language, so I actually understand the '/'. No prejudice to what you mentioned elsewhere though! I'll use 'and' and 'or' instead.

test/testenv.py
BrunoMSantos: Nice idea about the enum. Otherwise, yes, I'd rather make it explicit, especially as ultimately there should be no reason to presume or guess that one or the other is the default other than historic reasons.

README.rst
jnikula: Obviously C/C++ is widely used, it's just a kind of pet peeve of mine, and personally I avoid it when I can.

What to reply when someone asks, "do you know C/C++?"... "Very well/not so much."

Or, "do you like C/C++"... "Yes/no." :)

Anyway, I'll live with some C/C++ mentions to avoid repetition as long as it's not the main thing. Fair?

hawkmoth/parser.py
BrunoMSantos: > Does this not document the children of `UNEXPOSED_DECL` twice? For the first time as the children of `UNEXPOSED_DECL`, and the second time as the children of `tu.cursor.get_children()`?

No. The 2nd loop only recurses onto cursors that have a docstring associated with them and some cursors never get docstrings of their own according to our skip rules. `UNEXPOSED_DECL` is one of them and any preceding docstring is treated as a top level comment.

In fact I'm demoing this in some actual code now and one issue I have is kind of related: nothing shows up in the output because I'm not looking into namespaces, which will probably require a similar handling here.

README.rst
BrunoMSantos: Ahah, no I get that and to a large extent I agree. I'd say that this particular line might be one of the few places where it's less questionable, another one would be when talking about a C/C++ compiler and so on. I.e. it really means either in the most interchangeable sense. But I agree with changing them all.

Well, I wonder if it may be helpful to leave a few for the search engines xD

test/variable-cpp.stderr
```diff
@@ -0,0 +1,3 @@
+ERROR: variable.c:34: default initialization of an object of const type 'const char *(*const)(const char *)'
+ERROR: variable.c:39: default initialization of an object of const type 'const char *(*const volatile)(const char *)'
+ERROR: variable.c:44: default initialization of an object of const type 'const char *(*const *volatile *const[12][3])(const char *)'
```
jnikula, line 3: Here's something weird. I need to do this to make the tests pass:

```
diff --git a/test/variable-cpp.stderr b/test/variable-cpp.stderr
index 91db546cb779..681c232bf628 100644
--- a/test/variable-cpp.stderr
+++ b/test/variable-cpp.stderr
@@ -1,3 +1,3 @@
 ERROR: variable.c:34: default initialization of an object of const type 'const char *(*const)(const char *)'
 ERROR: variable.c:39: default initialization of an object of const type 'const char *(*const volatile)(const char *)'
-ERROR: variable.c:44: default initialization of an object of const type 'const char *(*const *volatile *const[12][3])(const char *)'
+ERROR: variable.c:44: default initialization of an object of const type 'const char *(*const *volatile *const [12][3])(const char *)'
```

i.e. add space.

I assume this works for you and looks like in github actions as well.

Clang 11 here.

jnikula: Clang 14 in github actions:

```
The following additional packages will be installed:
  libclang-14-dev python3-clang-14
```

test/variable.c
BrunoMSantos: True, I thought we tested the `restrict` path in other tests already, which is not the case apparently.

My suggestion would be to keep the change here, add other cases for various forms of `restrict` that compile in C++ as well and have a smaller test case for the C domain only with this particular case if you think the combination of features is relevant. Though I don't understand what crucial aspect makes it C specific so I don't know what to call it...

By the way, this prompted me to check the standards and I still don't understand the difference between C and C++. I'd need to change the commit message at least (for no discernable reason, I thought `restrict` was limited to function arguments and was wrongly questioning the validity of the snippet). Do you have any guesses as to why it won't work in C++?

FYI, this is what I get:
```
bruno@bms-tp ~ $ g++ a.cpp 
a.cpp:1:40: error: expected ‘)’ before ‘legal_type_involving_function_pointer’
    1 | const char* (*const *volatile *restrict legal_type_involving_function_pointer[12][3])(const char *in);
      |             ~                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                        )

bruno@bms-tp ~ $ clang++ a.cpp 
a.cpp:1:41: error: expected ')'
const char* (*const *volatile *restrict legal_type_involving_function_pointer[12][3])(const char *in);
                                        ^
a.cpp:1:13: note: to match this '('
const char* (*const *volatile *restrict legal_type_involving_function_pointer[12][3])(const char *in);
            ^
1 error generated.
```

test/variable-cpp.stderr
BrunoMSantos: > something to consider when refactoring the test folder?

Speaking of which, I had a look at it earlier and felt so lost I didn't get anywhere... I kind of gave up for now and returned to other issues :/

test/variable.c
BrunoMSantos: So, on a 2nd look, `restrict` does not seem to be part of C++ at all, but both gcc and clang accept them in some cases (no-op or not I couldn't say). Even this will not work in C++:
```c
int foo(int *restrict bar);
```

This only makes the particular case here weirder though, because this works in C++ and is conceptually the same albeit a much simpler version of the snippet that triggered this commit:
```c
int *restrict foo;
```

Anyway, I think I have an acceptable approach by putting all `restrict` stuff in its own test unit.

test/variable-cpp.stderr
jnikula: Yeah, just forget about the test folder stuff for now. I had a look too. It's a bit of a mess.

We define tests with .yaml, and link to a bunch of stuff via that, but then we expect .rst and .stderr to be there with the same basename. Arguably everything should be referenced via .yaml, and not magically assumed to be there with the same basename. It's nice to move stuff around when you don't have to worry about implicit stuff like that. OTOH that would be pretty verbose in the .yaml. Dunno.

jnikula: As to the Clang version issue, in this case the right fix might be to just fix the issue that is being warned about. I general, I'd prefer having the tests warning free apart from tests that test the warnings.

jnikula: It might also be the *easiest* fix.

BrunoMSantos: > It might also be the _easiest_ fix.

Good point, should have started there now that I think about it.

> We define tests with .yaml, and link to a bunch of stuff via that, but then we expect .rst and .stderr to be there with the same basename. Arguably everything should be referenced via .yaml, and not magically assumed to be there with the same basename. It's nice to move stuff around when you don't have to worry about implicit stuff like that. OTOH that would be pretty verbose in the .yaml. Dunno.

I'd like the explicit route actually. It's the easiest to use and, factoring the duplicated rst files, it should actually yield lower verbosity overall.

jnikula: > Good point, should have started there now that I think about it.

I fiddled with it a bit, and while I like to think I'm pretty good at C, I wasn't able to come up with an initializer for this:

```
const char* (*const *volatile *restrict legal_type_involving_function_pointer[12][3])(const char *in);
```

Maybe it should be split, idk.

BrunoMSantos: Oh, I fixed that locally already. Takes a very small piece of boilerplate, it's fairly obvious once you see it I think.

I did reduce the array size so that I didn't need to create a huge 12x3 matrix. And I still changed the `restrict` to `const` here so I can use this file for C++, but the boilerplate would be very similar either way.

I created a separate 'restrict' test unit that will be C only. Since `restrict` is not standard C++, I don't want to be testing those dark compiler corners in our tests.

I'll try to push all the fixes after dinner.

hawkmoth/__main__.py
BrunoMSantos: Ok, fixed _most_ of these. Left a total of 4, all in the documentation and in similar contexts, in places that somehow felt weird any other way.

You'll let me know if there are still too many :P

PR #117, opened 2023-01-11
C++ initial support v2

BrunoMSantos:
V2 of #116. There are a ton of small changes and some bigger ones.

Notably (apologies if I forget any major change):
- test: mass addition of cpp domain tests                          
  - Files renamed to somehow control the mayhem. Didn't do the same for C tests as that would entail too much noise.
- test: add default compiler options according to the test domain  
  - Optional. I'm not 100% confident it's an improvement, but it does reduce verbosity in the yaml files.
  - If dropped, the cpp-* tests will be broken until the clang options are fixed accordingly.
- test: fix test so it produces no error messages                  
  - Fixes the issue with older Clang producing different white space in some error messages.
- test: fix C test that breaks C++ compilation                     
  - Separates `restrict` stuff into its own test.
- parser: add mixed domain support                                                             
  - New function handles blocks that are always undocumented in a more obvious way.
  - Right now this is only the `extern "C"`, but will be helpful in handling namespaces in C++ as well. I should say I didn't test it yet, but I'm quite confident it will go that way.
  - This will still ignore blocks that could be documented and the user elects not to. E.g. documented members of a structure that has no docstring itself will be skipped.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -438,7 +438,7 @@ def _parse_undocumented_block(domain, comments, errors, cursor, nest):
     return ret
 
 # Parse a file and return a tree of docstring.Docstring objects.
-def parse(filename, domain='c', clang_args=None):
+def parse(filename, domain=None, clang_args=None):
     errors = []
     index = Index.create()
 
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -135,11 +135,13 @@ def _comment_extract(tu):
     comments = {}
     current_comment = None
 
+    is_doc = lambda cursor: cursor and docstring.Docstring.is_doc(cursor.spelling)
+
     for token in tu.get_tokens(extent=tu.cursor.extent):
         # Handle all comments we come across.
         if token.kind == TokenKind.COMMENT:
             # If we already have a comment, it wasn't related to another cursor.
-            if current_comment and docstring.Docstring.is_doc(current_comment.spelling):
+            if is_doc(current_comment):
                 top_level_comments.append(current_comment)
             current_comment = token
             continue
@@ -156,14 +158,22 @@ def _comment_extract(tu):
                                  CursorKind.MACRO_INSTANTIATION]:
             continue
 
+        # Cursors that are 1) never documented themselves, and 2) not allowed
+        # between the comment and the actual cursor being documented.
+        if token_cursor.kind in [CursorKind.UNEXPOSED_DECL]:
+            if is_doc(current_comment):
+                top_level_comments.append(current_comment)
+            current_comment = None
+            continue
+
         # Otherwise the current comment documents _this_ cursor. I.e.: not a top
         # level comment.
-        if current_comment and docstring.Docstring.is_doc(current_comment.spelling):
+        if is_doc(current_comment):
             comments[token_cursor.hash] = current_comment
         current_comment = None
 
     # Comment at the end of file.
-    if current_comment and docstring.Docstring.is_doc(current_comment.spelling):
+    if is_doc(current_comment):
         top_level_comments.append(current_comment)
 
     return top_level_comments, comments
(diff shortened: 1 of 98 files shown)


hawkmoth/parser.py
```diff
@@ -387,6 +397,46 @@ def _clang_diagnostics(diagnostics, errors):
         errors.append(ParserError(ErrorLevel(diag.severity), filename,
                                   diag.location.line, diag.spelling))
 
+def _parse_undocumented_block(domain, comments, errors, cursor, nest):
+    """Parse undocumented blocks.
+
+    Some blocks define plenty of children that may be documented themselves
+    while the parent cursor itself has no documentation. One such example is the
+    `extern "C"` block.
+    """
```
BrunoMSantos, line 406: > - parser: add mixed domain support
>   - Right now this is only the `extern "C"`, but will be helpful in handling namespaces in C++ as well. I should say I didn't test it yet, but I'm quite confident it will go that way.

And I have spoken too soon of course :/

The issue with namespaces is indeed very similar. But there's an issue of recursion (namespaces within namespaces) that will require a rethink on some of these things. A bit more detail: proper recursion so far implies that there is a docstring on the parent. This commit introduces  a special case to peel off one parent layer that is not documented, but we actually need this inside the recursive parser to be truly generic.

That said, I'd like to have less baggage to rebase as I move forward, so I'd ask for some leniency there if it can be spared. Comments welcome of course, but maybe that could be part of a series to add namespace support.

The plan would go:
1. initial support
2. C++ types support (cpp-class-wip being a preview)
3. namespaces
4. fixes that will surely be needed

hawkmoth/parser.py
```diff
@@ -316,7 +388,7 @@ def _clang_diagnostics(diagnostics, errors):
                                   diag.location.line, diag.spelling))
 
 # Parse a file and return a tree of docstring.Docstring objects.
-def parse(filename, clang_args=None):
+def parse(filename, domain='c', clang_args=None):
```
jnikula: I realize that "domain" is a Sphinx thing, and this sort of "leaks" that into the parser, which I try to avoid.

I'm going to ignore it though, because it's not that bad in this case, and I don't want to force another respin. And I don't even know yet what it would look like, and it's easier to fix afterwards, if indeed it needs to be fixed at all.

I'm thinking maybe it should just be "language" in the parser level, and maybe it could be an enum. And then it would be translated to domain in docstring. Anyway, not worth it at this time, just musing.

hawkmoth/parser.py
jnikula: In fact, I was going to comment that even `extern "C"` and `extern "C++"` can be nested!

And I was going to say it's fine, for now.

The `cursor.hash in comments` will have to move inside `_recursive_parse()` I guess.

BrunoMSantos: > In fact, I was going to comment that even extern "C" and extern "C++" can be nested!

Oh, I didn't know that. That's a good test case, but there's no support for it right now of course :/

I wonder if this commit (and the relevant test case) should be left out of the series then. We know this has a solution and this will remain unusable for C++ until we handle namespaces anyway. Fixing this after or when we fix the namespaces will be trivial in principle.

The rest of the series should hold up without it, though I'd have to confirm.

BrunoMSantos:
> I opted to not merge the documentation commit yet, because on the one hand I had some nitpicks, and on the other hand I would have wanted to go through it with more thought as well. But I didn't want to block the entire thing any longer, and it's separate from the rest anyway.
> 
> This I merged everything else on the command line. Many thanks for the work, and I look forward for more!

Ah my other reply didn't reach you in time. Well, no biggie. Many thanks for your patience and reviews :)

I confess I'm less optimistic now we'll pick this for our project due to risks and timeline, but that was always a possibility. I'll show a demo tomorrow and we'll see.

I'll carry this through either way though! I've invested too much already and I don't want another such opportunity to be missed.

hawkmoth/parser.py
BrunoMSantos: Good point. Will keep in mind :)

jnikula:
> I confess I'm less optimistic now we'll pick this for our project due to risks and timeline, but that was always a possibility. I'll show a demo tomorrow and we'll see.

Good luck, fingers crossed!

Remember to mention on the plus side that you know how it all works, and the maintainer is supportive. ;)

BrunoMSantos:
> I opted to not merge the documentation commit yet, because on the one hand I had some nitpicks, and on the other hand I would have wanted to go through it with more thought as well.

By the way, I'd also be interested in your thoughts on that when you have the time. I confess I was going for the bare minimum, just to say that it was there and maybe prompt experimentation and bug reports / feature requests.

Anyway, I can keep pushing actual C++ functionality without docs for now. It truly doesn't matter until we announce support publicly.

Actually I have the next batch _almost_ ready. It's a decent one I think :)

Preview in my `cpp-types` branch. I want to add `class enum` support before doing a new PR though. And as far as I can see, the only obvious thing missing will be namespaces and reworking that recursiveness thing.

jnikula:
> By the way, I'd also be interested in your thoughts on that when you have the time. I confess I was going for the bare minimum, just to say that it was there and maybe prompt experimentation and bug reports / feature requests.
> 
> Anyway, I can keep pushing actual C++ functionality without docs for now. It truly doesn't matter until we announce support publicly.

Part of the my concern was claiming C++ support in the main README that shows up in the github landing page long before we have a release with any support. I might postpone that update until *after* there's a release.

jnikula:
Oh, how'd the demo go, if you don't mind me asking? Curious and all. :)

jnikula:
> Points against (other than risk) is that comments are currently in Doxygen despite being the 'easy' subset of Doxygen and relatively small code base so far. I showed them the transformations stuff, but I also mentioned all the caveats.

FWIW I'm pretty far at making transformations Sphinx extensions, and adding Javadoc as a built-in extension. Feature wise, the conversion is the same and as limited as before, though, and Doxygen does have more syntax than Javadoc.

With the transformations moved to extensions, and out of the core code base, I'd actually be more open to taking contributions making them more feature rich, and even adding a Doxygen built-in extension. However, I'm pretty strongly in favour of just using plain rst for documentation comments (you know, Sphinx and all), so I have no intention of working on improving them myself, besides aiming for feature parity with what we have now.

And, of course, you can write extensions for this yourself, without waiting them to be merged to Hawkmoth.

One thing that I don't think I'll ever want to add is the Doxygen after-member comments with `/**< comment */` (https://www.doxygen.nl/manual/docblocks.html#memberdoc). 

I also think the proliferation of different accepted styles to add documentation comments is counter-productive, but I might be open to identifying and unwrapping the documentation comments a Sphinx event and extensible. That's a very long-term idea though.

Just opening up my views on this, so there are no surprises.

BrunoMSantos:
> One thing that I don't think I'll ever want to add is the Doxygen after-member comments with `/**< comment */` (https://www.doxygen.nl/manual/docblocks.html#memberdoc).

Hmm, that one has crossed my mind a few times, though, to be fair, I don't remember the last time I found a strong use case for it. Also why I never bothered so far I guess.

I'll try to convince you with code if it ever comes up :P

jnikula:
> > One thing that I don't think I'll ever want to add is the Doxygen after-member comments with `/**< comment */` (https://www.doxygen.nl/manual/docblocks.html#memberdoc).
> 
> Hmm, that one has crossed my mind a few times, though, to be fair, I don't remember the last time I found a strong use case for it. Also why I never bothered so far I guess.
> 
> I'll try to convince you with code if it ever comes up :P

Heh. I'm just leaning more towards "here's a way to do things nicely and uniformly, use it" instead of "here are ten different ways of doing things, anything goes, go wild, mix and match in the same project".

PR #118, opened 2023-01-11
Miscelaneous changes

BrunoMSantos:
Some changes I have rebased a few too many times today.

The gitignore is clearly a leftover that was previously missed. Don't think it will be contentious.

The other one is a very convenient target when checking what goes wrong on a given test. Maybe there are better workflows though? I'd be glad for it to be dropped in that case. Even if that is not the case, I'm still ok with keeping it as a local change.

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,6 +1,0 @@
-# Note: needs fixing if non-generated .script files get added
-*.script
-*.stderr-actual
-*.stderr-diff
-*.stdout-actual
-*.stdout-diff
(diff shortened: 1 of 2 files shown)


jnikula:
Wondering if it should be `verbose-test` or whether `quick-test` should be renamed. *shrug*. No big deal.

BrunoMSantos:
Great, now I can't unsee it :P

I've had that target locally for ages, since before the quick test (which I never use). I never questioned the name.

Thanks for the merge!

PR #119, opened 2023-01-11
Test updates

jnikula:
A couple of fixes, and a bunch of changes to specify various files explicitly via the testcase yaml instead of deducing them from the filename. It's not quite enough to make it work for for subdirectories, but it's a good step in that direction.

@BrunoMSantos There's no rush with reviewing this one, I know you've got other priorities right now. More like FYI.

### DIFF
--- a/test/autoenum.yaml
+++ b/test/autoenum.yaml
@@ -6,4 +6,4 @@ directive-options:
   file: enum.c
   members:
     - bar
-
+expected: autoenum.rst
--- a/test/autostruct-no-members.yaml
+++ b/test/autostruct-no-members.yaml
@@ -4,3 +4,4 @@ directive-arguments:
   - sample_struct
 directive-options:
   file: struct.c
+expected: autostruct-no-members.rst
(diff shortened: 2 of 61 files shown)


BrunoMSantos (approved):
It's an easy review this one: it's good to go.

I would only add one more commit: now we could reuse most, if not all, the .rst files from the C cases for the C++ cases and maybe some of the .stderr ones.

It can be separate to the series though.

jnikula:
> I would only add one more commit: now we could reuse most, if not all, the .rst files from the C cases for the C++ cases and maybe some of the .stderr ones.

There aren't all that many exact dupes though. The domain is different. Unless we want to go for running format on the entire files, and have `{domain}` in there. But that's definitely for another time.

BrunoMSantos:
> > I would only add one more commit: now we could reuse most, if not all, the .rst files from the C cases for the C++ cases and maybe some of the .stderr ones.
> 
> There aren't all that many exact dupes though. The domain is different. Unless we want to go for running format on the entire files, and have `{domain}` in there. But that's definitely for another time.

Ah, of course. Not a bad idea to support templating in expected results eventually though!

PR #120, opened 2023-01-13
C++ type support

BrunoMSantos:
So this is a big one feature wise, but I think it turned out quite neat for all it does.

The bottom line: this adds support for classes and class like structs and enums; templates; and even improves on extraction of qualifiers and weird specifiers that I think are relevant enough API wise (constexpr, mutable, override... the works). I'm honestly unaware of any single current C++ feature that is not covered within this scope. Hope I'm not wrong.

The one (only?) conspicuous absence here is namespace traversal and tracking. Granted that's a huge one which basically renders all of this pretty much useless, but still nice. Also, I wouldn't dare try to solve namespaces in this series.

Implementation wise, some of these things required me to drop to the token level, which is never ideal, but I tried to do so sparingly and only for relevant cursors to keep the performance hit to a minimum. There are a few places where small legibility improvements can be had perhaps if we loose some consistency in the helper functions' return values (returning None or '') and some function / variable names took a while to pick and I'm still unsure I picked the right ones.

Greatest pet peeve of mine right now is found early in the series with the addition of a `quals` argument to `Docstring`. These are meant for very specific qualifiers, but without changing too much in docstring (and perhaps the parser), I don't see an easy way around something like that. Maybe we can name it `method_quals` instead, but it's a mouthful and doesn't help that much. My best idea so far, but I would prefer to explore that separately, would be a `kwargs` approach to docstring and offload more of the formatting to individual specializations of docstring or other common bases for them instead of trying to cram so many different things into a single rigid API.

No attempt was given at updating the docs.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -321,6 +321,18 @@ def _specifiers_fixup(cursor, basetype):
 
     return type_elem
 
+def _get_scopedenum_type(cursor):
+    """Get the explicit underlying type of a scoped enumerator.
+
+    Returns:
+        Underlying type of a scoped enumerator that has been explicitly defined.
+        ``None`` otherwise.
+    """
+    if cursor.kind == CursorKind.ENUM_DECL and cursor.is_scoped_enum():
+        if list(cursor.get_tokens())[3].spelling == ':':
+            return f': {cursor.enum_type.spelling}'
+    return None
+
 def _type_fixup(cursor):
     """Fix non trivial types' spelling and append qualifiers.
 
@@ -389,8 +401,15 @@ def _type_fixup(cursor):
         name = cursor.spelling + dims
 
     inheritance = _get_inheritance(cursor)
+    scopedenum_type = _get_scopedenum_type(cursor)
 
-    return ttype, f'{name}{inheritance if inheritance else ""}'
+    colon_suffix = ''
+    if inheritance:
+        colon_suffix = inheritance
+    if scopedenum_type:
+        colon_suffix = scopedenum_type
+
+    return ttype, f'{name}{colon_suffix}'
 
 def _get_args(cursor):
     """Get function / method arguments."""
@@ -514,9 +533,14 @@ def _recursive_parse(domain, comments, errors, cursor, nest):
                                           nest=nest, name=name,
                                           decl_name=decl_name, meta=meta)
         elif cursor.kind == CursorKind.ENUM_DECL:
-            ds = docstring.EnumDocstring(domain=domain, text=text,
-                                         nest=nest, name=name,
-                                         decl_name=decl_name, meta=meta)
+            if cursor.is_scoped_enum():
+                ds = docstring.EnumClassDocstring(domain=domain, text=text,
+                                                  nest=nest, name=name,
+                                                  decl_name=decl_name, meta=meta)
+            else:
+                ds = docstring.EnumDocstring(domain=domain, text=text,
+                                             nest=nest, name=name,
+                                             decl_name=decl_name, meta=meta)
         elif cursor.kind in [CursorKind.CLASS_DECL, CursorKind.CLASS_TEMPLATE]:
             ds = docstring.ClassDocstring(domain=domain, text=text,
                                           nest=nest, name=name,
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -233,6 +233,23 @@ def _get_storage_class(cursor):
 
     return storage_class_map.get(cursor.storage_class)
 
+def _get_fn_quals(cursor):
+    """Get all the qualifiers of a function object.
+
+    Returns:
+        List of (prefix) function qualifiers.
+    """
+    fn_quals = []
+
+    if cursor.kind == CursorKind.FUNCTION_DECL:
+        tokens = [t.spelling for t in cursor.get_tokens()]
+        if 'static' in tokens:
+            fn_quals.append('static')
+        if 'inline' in tokens:
+            fn_quals.append('inline')
+
+    return fn_quals
+
 def _type_fixup(cursor):
     """Fix non trivial types' spelling and append qualifiers.
 
@@ -398,7 +415,8 @@ def _recursive_parse(domain, comments, errors, cursor, nest):
 
     elif cursor.kind == CursorKind.FUNCTION_DECL:
         args = _get_args(cursor)
-        ttype = cursor.result_type.spelling
+        linkage = ' '.join(_get_fn_quals(cursor))
+        ttype = f'{linkage}{" " if linkage else ""}{cursor.result_type.spelling}'
 
         ds = docstring.FunctionDocstring(domain=domain, text=text,
                                          nest=nest, name=name,
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -420,7 +420,8 @@ def _recursive_parse(domain, comments, errors, cursor, nest):
 
         ds = docstring.FunctionDocstring(domain=domain, text=text,
                                          nest=nest, name=name,
-                                         ttype=ttype, args=args, meta=meta)
+                                         ttype=ttype, args=args,
+                                         quals='', meta=meta)
         return [ds]
 
     # If we reach here, nothing matched i.e. there's a documentation comment
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -293,6 +293,57 @@ def _get_access_specifier(cursor):
 
     return name_map.get(cursor.access_specifier, None)
 
+def _get_template_line(cursor):
+    """Get the template arguments of a cursor.
+
+    This recurses for templated template arguments.
+
+    Returns:
+        String with the form 'template<...> ' if the cursor is a template or
+        `None` otherwise. When the cursor represents a templated template
+        argument, the returned string is actually of the form 'template<...>
+        name', but this should only occur under recursion.
+    """
+    # We only add the name when we recurse, in which case we need to track the
+    # name of the templated template argument. Otherwise the name is not part of
+    # the template arguments.
+    name = ''
+
+    if cursor.kind not in [CursorKind.CLASS_TEMPLATE,
+                           CursorKind.FUNCTION_TEMPLATE,
+                           CursorKind.TEMPLATE_TEMPLATE_PARAMETER]:
+        return None
+
+    # The type of type parameters can be 'typename' and 'class'. These are
+    # equivalent, but we want it to look like the source code for consistency.
+    # We can do it by looking at the tokens directly. This is slightly
+    # complicated due to variadic template type parameters.
+    def typetype(cursor):
+        tokens = list(cursor.get_tokens())
+        if tokens[-2].spelling == '...':
+            return f'{tokens[-3].spelling}...'
+        else:
+            return f'{tokens[-2].spelling}'
+
+    # We need to add the keyword 'typename' or 'class' if we have recursed and
+    # therefore we are inside the template argument list.
+    if cursor.kind == CursorKind.TEMPLATE_TEMPLATE_PARAMETER:
+        name = f' {typetype(cursor)} {cursor.spelling}'
+
+    template_args = []
+    for child in cursor.get_children():
+        if child.kind == CursorKind.TEMPLATE_TYPE_PARAMETER:
+            template_args.append(f'{typetype(child)} {child.spelling}')
+        elif child.kind == CursorKind.TEMPLATE_NON_TYPE_PARAMETER:
+            arg_name = f' {child.spelling}' if child.spelling != '' else '...'
+            template_args.append(f'{child.type.spelling}{arg_name}')
+        elif child.kind == CursorKind.TEMPLATE_TEMPLATE_PARAMETER:
+            arg = _get_template_line(child)
+            if arg:
+                template_args.append(arg)
+
+    return f'template<{", ".join(template_args)}>{name}'
+
 def _specifiers_fixup(cursor, basetype):
     """Fix the type for C++ specifiers.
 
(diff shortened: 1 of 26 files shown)


hawkmoth/parser.py
```diff
@@ -271,9 +286,18 @@ def _type_fixup(cursor):
...
+        if storage_class:
+            type_elem.append(storage_class)
+
+        type_elem.append(cursor_type.spelling)
+
         if stars_and_quals:
-            ttype += ' ' + stars_and_quals
+            type_elem.append(stars_and_quals)
+
+        ttype = ' '.join(type_elem)
```
jnikula, line 451: The thing that bugs me about this is that the documentation will be different depending on whether the documentation comments are above declarations or definitions. Declarations will have extern, definitions won't.

I think it's a matter of style whether one wants to put the documentation in the headers (declarations) or in the implementation (definitions). I don't think that should impact the resulting documentation.

I'd suggest not including `extern` here at all.

I think `static` is useful here, though.

test/function.c
```diff
@@ -1,12 +1,12 @@
...
  * Foo function.
  */
-int foo(int bar, int baz);
+static inline int foo(int bar, int baz);
 
 /**
  * No parameters.
  */
-int no_parameters(void);
+extern int no_parameters(void);
```
jnikula, line 9: Looking at the implementation, I'm not quite sure how we get `extern` for functions, but the same applies here as for variables. I'd prefer not having them. (Moreover, extern is superfluous for function declarations in C.)

I think `static` is useful. (Although documenting static functions would be more like documenting implementation details than an API, but it's still fine.)

I think `static inline` should mostly be an implementation detail of an API, but gut feeling says to err on the side of having it than trying to hide it, especially if we're going to have `static` anyway.

jnikula:
I read through it all at the high level. The `extern` stuff I commented on is something I'd like to have removed. Should not have a big impact, maybe a few lines at most.

I did not dig all that much into the details. I'll try to get into that later this weekend. Accessing the tokens so much obviously stands out as something of a disappointment, but if there's no other way, there's no other way.

jnikula:
I keep wondering about just passing the cursor to `Docstring` init. I think the current approach makes things unnecessarily difficult both sides of that interface. `parser.py` tries to shoehorn everything into those parameters, and `docstring.py` tries to convert them into useful things with `_fmt`. By passing the cursor, the whole thing would become much more object-oriented on the `Docstring` side, and a lot of the conversions could be more localized to relevant classes.

There's an additional thought as well. Originally, I had this idea of cross-checking e.g. the actual function parameters against the documented parameters, but I quickly realized I don't want to have anything to do with docstring parsing. However, we might want to allow extensions to do that. Pass the cursor and the docstring to a Sphinx event, and let extensions do that, while the core codebase stays clean.

Anyway, just throwing idea around. If you think passing cursors to `Docstring` makes things easier for you, I'm on board, but I'm definitely not asking you to do that. But I'll probably be more critical of modifying the `Docstring` interface to some *other* direction, although I'm not pre-emptively rejecting that either.

hawkmoth/parser.py
BrunoMSantos: > The thing that bugs me about this is that the documentation will be different depending on whether the documentation comments are above declarations or definitions. Declarations will have extern, definitions won't.
> 
> I think it's a matter of style whether one wants to put the documentation in the headers (declarations) or in the implementation (definitions). I don't think that should impact the resulting documentation.

Not just style. A user may want 2 sets of documentation: one for implementation details, one for the API users. Generally, I actually like having the distinction, though it is a difference from what we had been doing so far.

> I'd suggest not including `extern` here at all.

Hmm, if a header has an `extern` variable declared, that is API relevant I think. Of course, defining variables in a header is (almost) never a good idea so one could say the `extern` is redundant. But then there's that one case in which one needs to do just that. Not to mention I've seen a couple of cases where including `.c` files was the right thing to do...

> I think `static` is useful here, though.

test/function.c
BrunoMSantos: > Looking at the implementation, I'm not quite sure how we get `extern` for functions, but the same applies here as for variables. I'd prefer not having them. (Moreover, extern is superfluous for function declarations in C.)

We don't, I put it here to make it obvious that it doesn't show up in the documentation in this case. I agree, it's totally meaningless for functions unlike for variables.

> I think `static` is useful. (Although documenting static functions would be more like documenting implementation details than an API, but it's still fine.)
> 
> I think `static inline` should mostly be an implementation detail of an API, but gut feeling says to err on the side of having it than trying to hide it, especially if we're going to have `static` anyway.

Hmm, I think it conveys _something_ as an API, depending on user's expectations. Maybe some of this stuff could be made configurable one day if someone feels strongly about it.

BrunoMSantos:
> I keep wondering about just passing the cursor to `Docstring` init. I think the current approach makes things unnecessarily difficult both sides of that interface. `parser.py` tries to shoehorn everything into those parameters, and `docstring.py` tries to convert them into useful things with `_fmt`. By passing the cursor, the whole thing would become much more object-oriented on the `Docstring` side, and a lot of the conversions could be more localized to relevant classes.

I understand your idea and, to an extent, I like it. The only thing that keeps me wondering is whether I'd like to expose docstring to Clang's bindings directly, which are not the easiest thing to deal with as we're finding out more often than not (looking into tokens and whatnot).

This is something I've been thinking about too though and a similar idea (in the end result) that I like more though would be to derive our own version of a Clang `Cursor` that implements all the stuff we want in docstring and hides all of that craziness inside. Then top level parser only determines which specialization of docstring to call, constructs this new object out of the Clang cursor and hands it over.

> There's an additional thought as well. Originally, I had this idea of cross-checking e.g. the actual function parameters against the documented parameters, but I quickly realized I don't want to have anything to do with docstring parsing. However, we might want to allow extensions to do that. Pass the cursor and the docstring to a Sphinx event, and let extensions do that, while the core codebase stays clean.

I'm counting on having this feature honestly, one way or another. I'd still like the default of doing it for the limited case of Sphinx notation, maybe through an internal extension that kicks in when `:transform:` is `None`? Those using transformations could be left to their own pain :P

> Anyway, just throwing idea around. If you think passing cursors to `Docstring` makes things easier for you, I'm on board, but I'm definitely not asking you to do that. But I'll probably be more critical of modifying the `Docstring` interface to some _other_ direction, although I'm not pre-emptively rejecting that either.

BrunoMSantos:
> I read through it all at the high level. The `extern` stuff I commented on is something I'd like to have removed. Should not have a big impact, maybe a few lines at most.

Forgot about this one. Yeah, as you may have gathered by my other comments, I'd like to challenge that one :sweat_smile: 

It only shows up in variables, and I think it's relevant there. But let's see :)

BrunoMSantos:
> > There's an additional thought as well. Originally, I had this idea of cross-checking e.g. the actual function parameters against the documented parameters, but I quickly realized I don't want to have anything to do with docstring parsing. However, we might want to allow extensions to do that. Pass the cursor and the docstring to a Sphinx event, and let extensions do that, while the core codebase stays clean.
> 
> I'm counting on having this feature honestly, one way or another. I'd still like the default of doing it for the limited case of Sphinx notation, maybe through an internal extension that kicks in when `:transform:` is `None`? Those using transformations could be left to their own pain :P

Actually, presuming that all transformations must ultimately generate Sphinx notation, this check can (should) happen _after_ transformation, giving us universal parameter checking.

Whether we do so internally as a plugin or not I have no opinion as I currently know little of how that would work relative to other solutions.

jnikula:
> Actually, presuming that all transformations must ultimately generate Sphinx notation, this check can (should) happen _after_ transformation, giving us universal parameter checking.
> 
> Whether we do so internally as a plugin or not I have no opinion as I currently know little of how that would work relative to other solutions.

I think the main benefit of using plugins for this is that it's easy to experiment with them without committing to some solution in core code.

There are a number of things that could be plugins at docstring level, with the Sphinx extension providing callbacks to docstring that end up as events to extensions:

- Identifying what are documentation comments
- Removing comment markers
- Converting the cursor to a Sphinx domain directive
- Transforming documentation comment from one format to another
- Validating documentation comment contents

Many of these would actually benefit from having the cursor around in docstring level to be passed to the extensions.

Anyway, that's increasingly off-topic for the merge request at hand. :)

hawkmoth/parser.py
```diff
@@ -415,15 +481,16 @@ def _recursive_parse(domain, comments, errors, cursor, nest):
...
-        args = _get_args(cursor)
-        linkage = ' '.join(_get_fn_quals(cursor))
-        ttype = f'{linkage}{" " if linkage else ""}{cursor.result_type.spelling}'
-
+    elif cursor.kind in [CursorKind.FUNCTION_DECL,
+                         CursorKind.CONSTRUCTOR,
+                         CursorKind.DESTRUCTOR,
+                         CursorKind.CXX_METHOD,
+                         CursorKind.FUNCTION_TEMPLATE]:
+        ttype, args, quals = _fn_fixup(cursor)
```
jnikula, line 626: I'm thinking maybe we should keep `CursorKind.FUNCTION_DECL` handling separate from the rest, which are all C++ only, and not try to handle them all together. Bundling them leads to a lot of unnecessary stuff in `_fn_fixup()` for regular functions. Similarly, split `_get_fn_quals` to two.

AFAICT `_fn_fixup()` does not need the `CursorKind.FUNCTION_DECL` parts of `_get_fn_quals()`.

hawkmoth/parser.py
```diff
@@ -293,6 +293,34 @@ def _get_access_specifier(cursor):
...
+                       CursorKind.FIELD_DECL,
+                       CursorKind.PARM_DECL]:
+        tokens = [t.spelling for t in cursor.get_tokens()]
+
+        if 'mutable' in tokens:
+            type_elem.append('mutable')
+
+        if 'constexpr' in tokens:
+            type_elem.append('constexpr')
+            type_elem.append(basetype.spelling[len('const '):])
```
jnikula, line 369: Yikes that feels magic.

hawkmoth/parser.py
```diff
@@ -477,11 +495,15 @@ def _recursive_parse(domain, comments, errors, cursor, nest):
...
-                         CursorKind.ENUM_DECL]:
+    elif cursor.kind in [CursorKind.STRUCT_DECL,
+                         CursorKind.UNION_DECL,
+                         CursorKind.ENUM_DECL,
+                         CursorKind.CLASS_DECL,
+                         CursorKind.CLASS_TEMPLATE]:
 
         # Do not set the decl_name for anonymous symbols (empty spelling).
-        decl_name = name if cursor.spelling != '' else None
+        _, decl_name = _type_fixup(cursor)
```
jnikula, line 583: It scares me a bit that a lot of the paths now call `_type_fixup()` and I'm no longer quite sure why. It does a lot of things, and I'm completely uncertain if all of the things are needed in all of the paths.

Certainly the function pointer and array stuff is completely  unnecessary for this path.

I'm thinking smaller and more dedicated building blocks, so that changing `_type_fixup()` doesn't become insurmountable.

hawkmoth/parser.py
```diff
@@ -389,8 +401,15 @@ def _type_fixup(cursor):
...
+    scopedenum_type = _get_scopedenum_type(cursor)
 
-    return ttype, f'{name}{inheritance if inheritance else ""}'
+    colon_suffix = ''
+    if inheritance:
+        colon_suffix = inheritance
+    if scopedenum_type:
+        colon_suffix = scopedenum_type
+
+    return ttype, f'{name}{colon_suffix}'
```
jnikula: Same as previous patch, is it necessary to handle scoped enum type within `_type_fixup()`, or could it just be a separate thing for scoped enums?

jnikula:
For details, as I've mentioned in a few comments, I worry about `_fn_fixup()` and especially `_type_fixup()` combining a lot of stuff together, probably too much.

You've convinced me on the `extern` stuff. I also got confused by the `extern` being added in a function.

At the high level everything looks pretty good, it's just the fixup details I'd like to have cleaned up a bit.

hawkmoth/parser.py
BrunoMSantos: True, I was trying to keep all the common paths common, but gains are limited. I have a slight prior preference for the current version in this case, but let me see how it looks if I split it.

hawkmoth/parser.py
```diff
@@ -339,6 +380,31 @@ def _get_args(cursor):
 
     return args
 
+def _fn_fixup(cursor):
+    """Parse C++ method specific qualifiers."""
```
BrunoMSantos, line 487: This clearly fell behind.

hawkmoth/parser.py
BrunoMSantos: I actually wondered for a second or two whether that might be the case, but ultimately decided it was quite obvious. Good having someone checking my assumptions :)

So `constexpr` gets translated by Clang to `const` as part of the type spelling, which it is as the 1st implies the 2nd. However we want `constexpr` instead since it has a broader meaning than just `const`. We could leave it as `constexpr const`, but I think it looks a bit ugly and redundant. I'll add a comment to that effect.

hawkmoth/parser.py
BrunoMSantos: Yup, this one is less defensible. I'm only adding a couple of lines, but the function is already pretty wild. We need some parts, but I should be able to split it in a decent way.

hawkmoth/parser.py
BrunoMSantos: Yeah, like before, I agree with the sentiment despite a few common lines of code. I think we should be able to at least split new type declarations from variable / function argument declarations.

BrunoMSantos:
> For details, as I've mentioned in a few comments, I worry about `_fn_fixup()` and especially `_type_fixup()` combining a lot of stuff together, probably too much.
> 
> You've convinced me on the `extern` stuff. I also got confused by the `extern` being added in a function.
> 
> At the high level everything looks pretty good, it's just the fixup details I'd like to have cleaned up a bit.

Many thanks for the comments! I'll try to address them :)

BrunoMSantos:
Pushed a v2. The fixes were a bit tricky rebase wise and I ended up uncovering a couple of big issues along the way, though nothing fundamentally changed.

PR #122, opened 2023-01-17
C++ types v2

BrunoMSantos:
This supersedes #120.

Other than the proposed changes, I noticed and fixed some small issues:
1. Constructors / destructors were not getting an access specifier. Somehow it fell
through the cracks before, but it's now fixed and reflected in the test cases.
2. Member variables were not getting an access specifier. This was a big one, can't
imagine how I missed it, but it was even more flagrant than the c/dtors. Fixed
now. This is by far the biggest difference and is pretty much limited to the commit [parser: improve support for C++ specifiers](https://github.com/jnikula/hawkmoth/commit/d7f7005c51e85a6febfca043f76c6e7a1a4c80c1)
3. A few code paths were simplified. Partly due to the now clearer type fixup
thingy which made the improvements obvious.
4. Some mutually exclusive branches were improved for marginally better
performance: if one, avoid the other.

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -233,6 +233,23 @@ def _get_storage_class(cursor):
 
     return storage_class_map.get(cursor.storage_class)
 
+def _get_fn_quals(cursor):
+    """Get all the qualifiers of a function object.
+
+    Returns:
+        List of (prefix) function qualifiers.
+    """
+    fn_quals = []
+
+    if cursor.kind == CursorKind.FUNCTION_DECL:
+        tokens = [t.spelling for t in cursor.get_tokens()]
+        if 'static' in tokens:
+            fn_quals.append('static')
+        if 'inline' in tokens:
+            fn_quals.append('inline')
+
+    return fn_quals
+
 def _type_fixup(cursor):
     """Fix non trivial types' spelling and append qualifiers.
 
@@ -398,7 +415,8 @@ def _recursive_parse(domain, comments, errors, cursor, nest):
 
     elif cursor.kind == CursorKind.FUNCTION_DECL:
         args = _get_args(cursor)
-        ttype = cursor.result_type.spelling
+        linkage = ' '.join(_get_fn_quals(cursor))
+        ttype = f'{linkage}{" " if linkage else ""}{cursor.result_type.spelling}'
 
         ds = docstring.FunctionDocstring(domain=domain, text=text,
                                          nest=nest, name=name,
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -287,7 +287,7 @@ def _var_type_fixup(cursor):
             break
 
     if cursor_type.kind == TypeKind.FUNCTIONPROTO:
-        pad = lambda s: s if s.endswith('*') else s + ' '
+        pad = lambda s: s if s.endswith('*') or s.endswith('&') else s + ' '
 
         args = []
         for c in cursor.get_children():
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -233,22 +233,21 @@ def _get_storage_class(cursor):
 
     return storage_class_map.get(cursor.storage_class)
 
-def _get_fn_quals(cursor):
+def _get_function_quals(cursor):
     """Get all the qualifiers of a function object.
 
     Returns:
         List of (prefix) function qualifiers.
     """
-    fn_quals = []
+    tokens = [t.spelling for t in cursor.get_tokens()]
+    quals = []
 
-    if cursor.kind == CursorKind.FUNCTION_DECL:
-        tokens = [t.spelling for t in cursor.get_tokens()]
-        if 'static' in tokens:
-            fn_quals.append('static')
-        if 'inline' in tokens:
-            fn_quals.append('inline')
+    if 'static' in tokens:
+        quals.append('static')
+    if 'inline' in tokens:
+        quals.append('inline')
 
-    return fn_quals
+    return quals
 
 def _type_fixup(cursor):
     """Fix non trivial types' spelling and append qualifiers.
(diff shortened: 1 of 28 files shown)


hawkmoth/parser.py
```diff
@@ -385,7 +423,7 @@ def _get_args(cursor):
 
 def _function_fixup(cursor):
     """Parse additional details of a function declaration."""
-    args = _get_args(cursor)
+    args = _get_args(cursor, 'c')
```
jnikula, line 520: This "smells" a bit wrong, because `_function_fixup()` will also be called for C++. By reading the code I can see that it will work just fine with the 'c' domain, because the 'cpp' domain features within `_get_args()` are only relevant for methods etc.

But feels like the domain should be an invariant (apart from `extern "C"`) that you can rely on.

jnikula:
I might have some nitpicks here and there, but I'm not sure it's the best use of either of our time to respin this. It's probably easier to continue working on this after it's been merged, and I trust you to stick around. :wink: 

I found the coverage report in #123 to be useful for reference. In fact, this merge *increases* the coverage.

I also wrote a tool that takes a testcase `.yaml` and turns it into html, just so I could look at what the end results of the C++ domain look like, without turning them into examples. I'll polish and submit that at some point. Might prove useful for debugging later.

Anyway, are you happy with this yourself? Good to go?

jnikula:
Oh, on the nitpicks, some of the methods in the tests could use some more parameters. There are very few. But can be added later.

jnikula:
> I also wrote a tool that takes a testcase `.yaml` and turns it into html, just so I could look at what the end results of the C++ domain look like, without turning them into examples. I'll polish and submit that at some point. Might prove useful for debugging later.

Btw one of the things I noted was that Sphinx does not display public access specifier even though it's present in rst. Only protected/private.

BrunoMSantos:
> I might have some nitpicks here and there, but I'm not sure it's the best use of either of our time to respin this. It's probably easier to continue working on this after it's been merged, and I trust you to stick around. wink

I'm planning on it! Though, heads up, I will probably slow down a bit going forward. Everyone liked it and would ultimately prefer it to breathe, but with everything else that I have to do and without a clear decision from the team to make it work, I'm not comfortable enough taking on the risk myself. Besides nuclear fusion is delayed enough as it is xD

Instead, I'm planning on wiring up breathe and focus on other things. Once this is a bit more mature, I can attempt a conversion. Especially if breathe / Doxygen become annoying (I give ourselves good chances on that one).

Either way, I expect C++ will likely attract a lot more users and contributions than C did. However sad that is to me.

> I found the coverage report in #123 to be useful for reference. In fact, this merge _increases_ the coverage.

Well, if I mostly added code with ~100% coverage, relative coverage should increase even if uncovered lines didn't change much. Some sort of cognitive bias... :P

But it's a cool thing to put a number on. Also, really nice output, I can totally see a few missed cases, though I'd rather do a separate round to trim redundant code paths and add new tests after merging #123. Besides, it's not too bad right now!

> I also wrote a tool that takes a testcase `.yaml` and turns it into html, just so I could look at what the end results of the C++ domain look like, without turning them into examples. I'll polish and submit that at some point. Might prove useful for debugging later.

That's a neat idea. I often end up using our own docs to demo that, but it's a bit cumbersome.

> Anyway, are you happy with this yourself? Good to go?

Yeah, I'm pretty stoked with how it turned out so far and the previous round had bigger holes than this one for sure.

That said, I actually found one specifier I didn't take care of: `noexcept`. If we do a new round, it would be to add that, but rebasing this becomes increasingly costly and I was pretty sure I had caught everything before. We now know how that turned out.

> Oh, on the nitpicks, some of the methods in the tests could use some more parameters. There are very few. But can be added later.

Yeah, I may have been taking a bit of a white box approach to tests so far.

> Btw one of the things I noted was that Sphinx does not display public access specifier even though it's present in rst. Only protected/private.

Yeah, I noticed it. I think putting them there is the right behaviour though in case Sphinx changes. Also not sure if this can be affected by themes, haven't looked.

jnikula:
> > I might have some nitpicks here and there, but I'm not sure it's the best use of either of our time to respin this. It's probably easier to continue working on this after it's been merged, and I trust you to stick around. wink
> 
> I'm planning on it! Though, heads up, I will probably slow down a bit going forward. Everyone liked it and would ultimately prefer it to breathe, but with everything else that I have to do and without a clear decision from the team to make it work, I'm not comfortable enough taking on the risk myself. Besides nuclear fusion is delayed enough as it is xD
> 
> Instead, I'm planning on wiring up breathe and focus on other things. Once this is a bit more mature, I can attempt a conversion. Especially if breathe / Doxygen become annoying (I give ourselves good chances on that one).

Too bad, though not completely unexpected. All thing considered, it would have been a bit of a jump into the unknown, while Doxygen+Breathe is the safe choice. IMO it's over-complicated as a whole, especially for small projects, but then you can actually find support and solutions to the problems you hit.

> Either way, I expect C++ will likely attract a lot more users and contributions than C did. However sad that is to me.

Fingers crossed! :)

> > Anyway, are you happy with this yourself? Good to go?
> 
> Yeah, I'm pretty stoked with how it turned out so far and the previous round had bigger holes than this one for sure.
> 
> That said, I actually found one specifier I didn't take care of: `noexcept`. If we do a new round, it would be to add that, but rebasing this becomes increasingly costly and I was pretty sure I had caught everything before. We now know how that turned out.

On this one, up to you really!

> > Btw one of the things I noted was that Sphinx does not display public access specifier even though it's present in rst. Only protected/private.
> 
> Yeah, I noticed it. I think putting them there is the right behaviour though in case Sphinx changes.

Agreed.

BrunoMSantos:
> On this one, up to you really!

Let's merge it! :)

hawkmoth/parser.py
BrunoMSantos: Oh, missed this one, the notification went straight to the junk folder.

Yeah, you're right. Don't merge it yet then, I'll force push a fix for this later.

jnikula: I suggested merging *after* I'd commented on this one. :)

BrunoMSantos: Yeah, but this one made me feel particularly dirty xD

It's also the sort of thing that has no immediate pressure for a fix, which I don't like.

Anyway, I force pushed this one thing only. Thanks for the patience!

PR #123, opened 2023-01-18
test: add test coverage and reporting

jnikula:
@BrunoMSantos I wondered about the coverage of the tests, in particular the new C++ stuff, as well as what parts actually get executed of the all the things in parser.py.

I've looked at your #122 with this, and it's pretty good. Just run `make test-coverage-html` and browse the results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.pyc
 *~
+.coverage
 .venv
 build
 dist
 hawkmoth.egg-info
+test/coverage-report
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 flake8
 pytest
+pytest-cov
 pytest-xdist
 sphinx
 sphinx-testing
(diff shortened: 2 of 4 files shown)


BrunoMSantos:
This is great! Only thing I'd point out is that I'd prefer the reports to go on a non hidden folder, possibly inside the `test/` folder to mimic the documentation which uses `doc/_build`.

Anyway, I don't think it matters that much. Good to go.

jnikula:
> This is great! Only thing I'd point out is that I'd prefer the reports to go on a non hidden folder, possibly inside the `test/` folder to mimic the documentation which uses `doc/_build`.
> 
> Anyway, I don't think it matters that much. Good to go.

I noticed the `Coverage report: file:///path/to/hawkmoth/test/coverage-report/index.html` printout was wrong if not run from the top level directory, so fixed that and changed the report output directory to `test/coverage-report`.

PR #124, opened 2023-01-19
Testcase class

jnikula:
Another day, another test overhaul :)

This time, add a Testcase class that represents a testcase `.yaml`, and make `testcase` in code always refer to a Testcase object instead of a plain filename. Gradually convert most of the testcase handling into a nice class structure.

Some of the steps are a bit annoying diff-wise, but `git show -w` helps a bit.

A nice feature of the coverage reports: I could run coverage before and after this, compare the results, and observe there's no diff. :)

### DIFF
--- a/test/testenv.py
+++ b/test/testenv.py
@@ -88,14 +88,12 @@ class Testcase:
 
         return directive_str
 
-    def run_test(self, monkeypatch=None, capsys=None):
+    def run_test(self):
         if self.options.get('expected-failure'):
             pytest.xfail()
 
-        output_docs, output_errors = self.get_output(monkeypatch=monkeypatch,
-                                                     capsys=capsys)
-        expect_docs, expect_errors = self.get_expected(monkeypatch=monkeypatch,
-                                                       capsys=capsys)
+        output_docs, output_errors = self.get_output()
+        expect_docs, expect_errors = self.get_expected()
 
         assert output_docs == expect_docs
         assert output_errors == expect_errors
--- a/test/testenv.py
+++ b/test/testenv.py
@@ -49,6 +49,19 @@ class Testcase:
 
         return os.path.join(os.path.dirname(self.filename), relative)
 
+    def get_input_filename(self):
+        options = self.options
+
+        directive = options.get('directive')
+        if directive == 'autodoc':
+            arguments = options.get('directive-arguments', [])
+            basename = arguments[0]
+        else:
+            directive_options = options.get('directive-options', {})
+            basename = directive_options.get('file')
+
+        return self.get_relative_filename(basename)
+
 def get_testid(testcase):
     return testcase.get_testid()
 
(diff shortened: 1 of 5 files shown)


test/testenv.py
```diff
@@ -15,15 +15,10 @@
...
+    name = os.path.splitext(os.path.basename(testcase.filename))[0]
 
     return name
 
-def get_testcases(path):
-    for f in sorted(os.listdir(path)):
-        if f.endswith(testext):
-            yield os.path.join(path, f)
-
 options_schema = strictyaml.Map({
```
BrunoMSantos, line 16: This is missing a leading underscore, which is old news, but it could arguably be moved inside the class as well, which would make a nice pretext to rename.

test/testenv.py
```diff
@@ -43,6 +37,15 @@ def __init__(self, filename):
         with open(filename, 'r') as f:
             self.options = strictyaml.load(f.read(), options_schema).data
 
+    def get_testid(self):
+        """Convert a testcase filename into a test case identifier."""
+        name = os.path.splitext(os.path.basename(self.filename))[0]
```
BrunoMSantos, line 42: We're constructing some of these strings repeatedly. We could push some of this stuff to the constructor or something. Not critical performance wise though and it's certainly no worse than before.

test/testenv.py
```diff
@@ -43,6 +43,12 @@ def get_testid(self):
 
         return name
 
+    def get_relative_filename(self, relative):
```
BrunoMSantos, line 46: This ends up as an internal method only, so, respecting the naming convention, this should have a leading underscore. I'll leave it up to you on how and if you want to do that here though.

BrunoMSantos:
This looks pretty good at a glance, especially the latter commits where everything starts falling into place. I could nitpick a bit, but I also don't see much value in in it honestly.

I left few quick comments I stumbled on, but mostly because I had written them already. It's up to you really.

(By the way, not sure you had that in mind, but this should make it pretty easy to have sub folders too I think!)

test/testenv.py
jnikula: Pushed a commit fixing this.

test/testenv.py
jnikula: Pushed a commit fixing this.

test/testenv.py
jnikula: I'm not sure if I'll need it later, so I'll leave it as-is for now.

jnikula:
> (By the way, not sure you had that in mind, but this should make it pretty easy to have sub folders too I think!)

Forgot to mention in the description, but yeah, that was one of the reasons for doing this. I think it'll also help with the eventual switch from `sphinx-testing` to `sphinx.testing`, but this was just starting to grow so big I wanted to get it merged.

jnikula:
And as always, many thanks for the review!

BrunoMSantos:
Looks good to me!

PR #125, opened 2023-01-20
Add support for arranging tests in subdirectories

jnikula:
This should enable us to start arranging tests in subdirectories. I played with it a bit, and it seems to work, but I didn't actually include any rearrangement of the files. I threw in some additional asserts in case the files aren't found.

Examples work test-wise, but `make html` will fail if the examples are placed in a subdirectory, because of `hawkmoth_root` config. I think that's fine for now.

### DIFF
--- a/test/testenv.py
+++ b/test/testenv.py
@@ -101,9 +101,10 @@ def get_testid(testcase):
     return testcase.get_testid()
 
 def get_testcase_filenames(path):
-    for f in sorted(os.listdir(path)):
-        if f.endswith(testext):
-            yield os.path.join(path, f)
+    for root, dirs, files in sorted(os.walk(path)):
+        for f in files:
+            if f.endswith(testext):
+                yield os.path.join(root, f)
 
 def read_file(filename):
     if not filename or not os.path.isfile(filename):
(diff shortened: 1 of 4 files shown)


test/testenv.py
```diff
@@ -101,9 +101,10 @@ def get_testid(testcase):
     return testcase.get_testid()
```
BrunoMSantos, line 103: I'm not sure that the comment about the examples is warranted. `hawkmoth_root` simply has to be redirected to an examples folder and / or the directives must include the path of the files relative to `hawkmoth_root`. I.e. it _does_ work with respect to the changes we introduce here.

Anyway, nitpicking.

BrunoMSantos:
Looks pretty good to me!

test/testenv.py
jnikula: It'll work if you move all the examples to the same directory, and point `hawkmoth_root` there. But if you wanted to have examples in multiple directories, it'll break...

jnikula:
Thanks for the review, merged!

test/testenv.py
BrunoMSantos: My point is that this code is perfect as it is with regards to that. If we have `hawkmoth_root=test/examples`, we just need to do `.. c:autodoc:: dir1/dir2/src.c`, either way it's not a limitation with regards to what you changed here. We'd need only change `conf.py` and possibly adjust `update_examples.py` unless I've missed something.

Anyway, good merge ;)

PR #126, opened 2023-01-20
doc: fix intersphinx references

jnikula:
First, fix the intersphinx URL

Second, add 'confval' object type [1] to be able to reference objects of that type using intersphinx. (Yes, it's also possible to turn conf.py into an extension just by adding a setup() function.)

Third, fix the intersphinx references [2].

[1] https://github.com/sphinx-doc/sphinx/issues/5562#issuecomment-434296574 [2] https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

### DIFF
--- a/doc/extension.rst
+++ b/doc/extension.rst
@@ -18,7 +18,7 @@ dependencies.
 Usage
 -----
 
-Add ``hawkmoth`` to :data:`sphinx:extensions` in ``conf.py``. Note that
+Add ``hawkmoth`` to :confval:`sphinx:extensions` in ``conf.py``. Note that
 depending on the packaging and installation directory, this may require
 adjusting the :envvar:`python:PYTHONPATH`.
 
--- a/doc/extension.rst
+++ b/doc/extension.rst
@@ -4,9 +4,9 @@ C Autodoc Extension
 ===================
 
 Hawkmoth provides a Sphinx extension that adds :ref:`new directives
-<directives>` to the Sphinx :any:`C domain <sphinx:c-domain>` to incorporate
-formatted C source code comments into a document. Hawkmoth is Sphinx
-:any:`sphinx:sphinx.ext.autodoc` for C.
+<directives>` to the Sphinx :external+sphinx:ref:`C domain <c-domain>` to
+incorporate formatted C source code comments into a document. Hawkmoth is Sphinx
+:py:mod:`sphinx.ext.autodoc` for C.
 
 For this to work, the documentation comments must of course be written in
 correct reStructuredText. See :ref:`documentation comment syntax <syntax>` for
@@ -18,8 +18,8 @@ dependencies.
 Usage
 -----
 
-Add ``hawkmoth`` to :confval:`sphinx:extensions` in ``conf.py``. Note that
-depending on the packaging and installation directory, this may require
+Add ``hawkmoth`` to :external+sphinx:confval:`extensions` in ``conf.py``. Note
+that depending on the packaging and installation directory, this may require
 adjusting the :envvar:`python:PYTHONPATH`.
 
 For example:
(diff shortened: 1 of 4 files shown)


doc/extension.rst
```diff
@@ -4,9 +4,9 @@ C Autodoc Extension
 ===================
 
 Hawkmoth provides a Sphinx extension that adds :ref:`new directives
-<directives>` to the Sphinx :any:`C domain <sphinx:c-domain>` to incorporate
-formatted C source code comments into a document. Hawkmoth is Sphinx
-:any:`sphinx:sphinx.ext.autodoc` for C.
+<directives>` to the Sphinx :any:`C domain <c-domain>` to incorporate formatted
```
BrunoMSantos, line 7: I'd have preferred if these were a separate commit, if not in fact all '3' items in the message, but it's simple enough not to matter.

However, I don't see a reason to remove the explicit source of the tag. Looking at the docs I see no counter indication and explicit > implicit any time of day for me.

There's an `:external:` role as well (don't remember seeing that before), but that one is redundant and less specific than scoping these on each external link.

BrunoMSantos: Note that I don't know which part of the docs mentioned in the commit ([2]) is relevant here and I'm commenting under the assumption that it just isn't.

Locally the old references were good for me. Maybe a Sphinx v6 thing?

doc/extension.rst
```diff
@@ -18,9 +18,9 @@ dependencies.
 Usage
 -----
 
-Add ``hawkmoth`` to :data:`sphinx:extensions` in ``conf.py``. Note that
-depending on the packaging and installation directory, this may require
-adjusting the :envvar:`python:PYTHONPATH`.
+Add ``hawkmoth`` to :any:`extensions` in ``conf.py``. Note that depending on the
```
BrunoMSantos, line 21: Case in point, we actually have a reference called `extension` but we've been talking about having extensionS to hawkmoth ourselves... so it would be unsurprising if we aliased this one in the future.

I haven't tested it, but I believe intersphinx stops looking once it gets a hit, so we wouldn't even get a warning about the alias.

doc/extension.rst
jnikula: > I'd have preferred if these were a separate commit, if not in fact all '3' items in the message, but it's simple enough not to matter.

Yeah, I don't disagree. Usually it's a bad sign if the commit message contains lists of things (or a sentence beginning with "Also, ").

> However, I don't see a reason to remove the explicit source of the tag. Looking at the docs I see no counter indication and explicit > implicit any time of day for me.
> 
> There's an `:external:` role as well (don't remember seeing that before), but that one is redundant and less specific than scoping these on each external link.

The thing is, I couldn't find anything to back up our current usage of the explicit external references:

```
:any:`sphinx:foo`
```

Per the interpshinx docs the above should be:

```
:external+sphinx:any:`foo`
```

(The `any` role isn't significant here.)

I'm with you on being explicit rather than implicit, but that's just gets pretty verbose. I'm now leaning towards using it anyway, but I wonder where the current usage comes from, and why it works.

jnikula: I'll respin this, with explicit intersphinx references, but I don't know which style to go for.

doc/extension.rst
jnikula: Agreed.

doc/extension.rst
BrunoMSantos: > The thing is, I couldn't find anything to back up our current usage of the explicit external references:
> 
> ```
> :any:`sphinx:foo`
> ```
> 
> Per the interpshinx docs the above should be:
> 
> ```
> :external+sphinx:any:`foo`
> ```
> 
> (The `any` role isn't significant here.)
> 
> I'm with you on being explicit rather than implicit, but that's just gets pretty verbose. I'm now leaning towards using it anyway, but I wonder where the current usage comes from, and why it works.

Good points... I was also surprised I didn't see that syntax in the docs you linked, but didn't go hunting down why it was so. Maybe from an older version which is going out of style.

I'd still make the point that `external+sphinx` is better if we can't verify the old syntax, even if very verbose.

jnikula:
Force pushed v2 with the issues addressed.

BrunoMSantos:
> Force pushed v2 with the issues addressed.

Shame about the external syntax, but so be it. Looks good to me!

PR #127, opened 2023-01-20
parser: add missing 'noexcept' support

BrunoMSantos:
Before we forget... :)

### DIFF
--- a/hawkmoth/parser.py
+++ b/hawkmoth/parser.py
@@ -34,7 +34,7 @@ import enum
 from dataclasses import dataclass
 
 from clang.cindex import TokenKind, CursorKind, TypeKind
-from clang.cindex import StorageClass, AccessSpecifier
+from clang.cindex import StorageClass, AccessSpecifier, ExceptionSpecificationKind
 from clang.cindex import Index, TranslationUnit
 from clang.cindex import Diagnostic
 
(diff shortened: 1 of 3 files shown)


hawkmoth/parser.py
```diff
@@ -34,7 +34,7 @@
 from dataclasses import dataclass
 
 from clang.cindex import TokenKind, CursorKind, TypeKind
-from clang.cindex import StorageClass, AccessSpecifier
+from clang.cindex import StorageClass, AccessSpecifier, ExceptionSpecificationKind
```
jnikula, line 37: Side note, what do you think of this style of imports:

```
from clang.cindex import (AccessSpecifier,
                          CursorKind,
                          Diagnostic,
                          Index,
                          StorageClass,
                          TokenKind,
                          TranslationUnit,
                          TypeKind)

I think our cindex imports are getting a bit unweildy. Usually I'd prefer just `from clang import cindex` and then then using the `cindex.TokenKind` syntax everywhere, but here that gets too verbose in code, I think.

And absolutely off-topic for the change at hand!

```

BrunoMSantos: Yeah, I kind of like it. I agree with you too, not ideal, but too verbose in our case otherwise.

Thanks for the merge!

PR #128, opened 2023-01-21
GitHub actions update

jnikula:
Minor cleanups

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,18 +12,18 @@ jobs:
   build-docs:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - name: Install dependencies (apt)
-      run: sudo apt-get install -y python3-clang python3-pip
-    - name: Install dependencies (pip)
-      run: sudo pip install --requirement requirements.txt --upgrade --upgrade-strategy eager
-    - name: Build docs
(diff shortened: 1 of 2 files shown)


jnikula:
Ugh, some issues with specifying Python version and installing python3-clang, which can't be found when using Python 3.10. I give up.

BrunoMSantos:
Hmm, what's the default Python version?

When installing Clang through apt, it will most likely install it to a folder like `/lib/python3.10/site-packages/` (replace 3.10 by whatever Python version is default). Naturally that folder will only be in the search path for the respective version of Python.

Before considering ugly hacks, does it solve anything if we also add `clang` to the `requirements.txt` file? Presumably it will install the bindings for the right version and still be able to find `libclang` from the apt package.

BrunoMSantos:
Of course that the clang bindings installed that way are not 'official'...

By the way, I'd also contest the recommendation to specify the version honestly. v3.10 might be a bold thing to do today, it will be old and obsolete tomorrow... Presumably we're supporting this across multiple versions and we're not testing them all, why not use the Ubuntu default? It probably covers a lot more users anyway and would save us these troubles for when we have actual incompatibilities to consider.
Of course I'm biased though ahahah... I never enjoyed this increasingly common practice of multiple package managers and overly specific dependencies.

BrunoMSantos:
Looks good to me ;)

PR #129, opened 2023-01-24
Process docstring event

jnikula:
This series adds `hawkmoth-process-docstring` event for extending Hawkmoth, and converts all transformations to use it, turning them into built-in extensions.

### DIFF
--- a/doc/built-in-extensions.rst
+++ b/doc/built-in-extensions.rst
@@ -66,3 +66,39 @@ necessary. This may change in the future.
           'javadoc-liberal': doccompat.javadoc_liberal,
           'kernel-doc': doccompat.kerneldoc,
       }
+
+Converting to Event Handling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have a function ``foo_transform()`` that you use with
+:py:data:`cautodoc_transformations`, it can be used with the
+:event:`hawkmoth-process-docstring` event as follows in ``conf.py``.
+
+Before:
+
+.. code-block:: python
+
+   cautodoc_transformations = {
+       'foo': foo_transform
+   }
+
+After:
+
+.. code-block:: python
+
+   def _process_docstring(app, lines, transform, options):
+       if transform != 'foo':
+           return
+
+       comment = '\n.join(lines)
+       comment = foo_transform(comment)
+       lines[:] = comment.splitlines()[:]
+
+   # conf.py can be turned into an extension by adding setup() function
+   setup(app)
+       app.connect('hawkmoth-process-docstring', _process_docstring)
+
+Of course, if you modify ``foo_transform()`` to operate on a list of strings,
+you can do away with the ``join()`` and ``splitlines()`` pair. Also, this can be
+turned into a proper Sphinx extension by putting it in a separate package. See
+:external+sphinx:doc:`development/index` for details.
--- a/doc/built-in-extensions.rst
+++ b/doc/built-in-extensions.rst
@@ -6,6 +6,27 @@ Built-In Extensions
 Hawkmoth is :ref:`extensible <extending>`, and ships with some built-in
 extensions.
 
+.. _hawkmoth.ext.javadoc:
+
+hawkmoth.ext.javadoc
+--------------------
+
+This extension converts Javadoc_ comments to reStructuredText.
+
+.. _Javadoc: https://www.oracle.com/technetwork/java/javase/documentation/javadoc-137458.html
+
+Installation and configuration in ``conf.py``:
+
+.. code-block:: python
+
+   extensions.append('hawkmoth.ext.javadoc')
+
+.. py:data:: hawkmoth_javadoc_transform
+   :type: list
+
+   List of transforms to convert, defaults to ``['javadoc']``. If the
+   ``transform`` option is not in in this list, do nothing.
+
 .. _hawkmoth.ext.transformations:
 
 hawkmoth.ext.transformations
--- a/doc/built-in-extensions.rst
+++ b/doc/built-in-extensions.rst
@@ -1,0 +1,68 @@
+.. _built-in-extensions:
+
+Built-In Extensions
+===================
+
+Hawkmoth is :ref:`extensible <extending>`, and ships with some built-in
+extensions.
+
+.. _hawkmoth.ext.transformations:
+
+hawkmoth.ext.transformations
+----------------------------
+
+This extension handles the :py:data:`cautodoc_transformations` feature, using
+the :event:`hawkmoth-process-docstring` event.
+
+Going forward, it's recommended to handle transformations using the event
+directly instead of :py:data:`cautodoc_transformations`. This built-in extension
+provides backward compatibility for the functionality.
+
+Installation and configuration in ``conf.py``:
+
+.. code-block:: python
+
+   extensions.append('hawkmoth.ext.transformations')
+
+For now, this extension is loaded by default, and the above is not strictly
+necessary. This may change in the future.
+
+.. py:data:: cautodoc_transformations
+   :type: dict
+
+   Transformation functions for the :rst:dir:`c:autodoc` directive ``transform``
+   option. This is a dictionary that maps names to functions. The names can be
+   used in the directive ``transform`` option. The functions are expected to
+   take a (multi-line) comment string as a parameter, and return the transformed
+   string. This can be used to perform custom conversions of the comments,
+   including, but not limited to, Javadoc-style compat conversions.
+
+   The special key ``None``, if present, is used to convert everything, unless
+   overridden in the directive ``transform`` option. The special value ``None``
+   means no transformation is to be done.
+
+   For example, this configuration would transform everything using
+   ``default_transform`` function by default, unless overridden in the directive
+   ``transform`` option with ``javadoc`` or ``none``. The former would use
+   ``javadoc_transform`` function, and the latter would bypass transform
+   altogether.
+
+   .. code-block:: python
+
+      cautodoc_transformations = {
+          None: default_transform,
+          'javadoc': javadoc_transform,
+          'none': None,
+      }
+
+   The example below shows how to use Hawkmoth's existing compat functions in
+   ``conf.py``.
+
+   .. code-block:: python
+
+      from hawkmoth.util import doccompat
+      cautodoc_transformations = {
+          'javadoc-basic': doccompat.javadoc,
+          'javadoc-liberal': doccompat.javadoc_liberal,
+          'kernel-doc': doccompat.kerneldoc,
+      }
--- a/doc/built-in-extensions.rst
+++ b/doc/built-in-extensions.rst
@@ -27,6 +27,27 @@ Installation and configuration in ``conf.py``:
    List of transforms to convert, defaults to ``['javadoc']``. If the
    ``transform`` option is not in in this list, do nothing.
 
+
+.. _hawkmoth.ext.napoleon:
+
+hawkmoth.ext.napoleon
+---------------------
+
+This extension provides a bridge from Hawkmoth to the
+:external+sphinx:py:mod:`sphinx.ext.napoleon` extension.
+
+Installation and configuration in ``conf.py``:
+
+.. code-block:: python
+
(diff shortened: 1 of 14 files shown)


BrunoMSantos:
Sounds promising! Sorry I haven't picked it up yet, I'll review this between today / tomorrow.

doc/extension.rst
```diff
@@ -121,6 +81,14 @@ The extension has a few configuration options that can be set in ``conf.py``:
    You can also pass in the compiler to use, for example
    ``get_include_args('gcc')``.
 
+.. py:data:: cautodoc_transformations
```
BrunoMSantos, line 84: So this is the 2nd definition of this symbol in this commit (other one in `built-in-extensions`) and Sphinx dutifully complains about it. Haven't read through this carefully yet, but this looks like some sort of mistake.

doc/extension.rst
```diff
@@ -121,6 +81,14 @@ The extension has a few configuration options that can be set in ``conf.py``:
...
+   :type: dict
+   :noindex:
+
+   This configuration option and its functionality has been moved to the
+   built-in :ref:`hawkmoth.ext.transformations` extension. See
+   :py:data:`cautodoc_transformations`.
```
BrunoMSantos, line 90: On a similar note, is this supposed to be a reference to itself?

doc/extension.rst
jnikula: The `:noindex:` here should avoid the complaints, and prevent this from being a link target. Sometimes Sphinx caching causes an issue when things change; try `make clean`.

doc/extension.rst
jnikula: Same here, with `:noindex:` this should be a reference to the definition in `built-in-extensions.rst`.

doc/built-in-extensions.rst
```diff
@@ -66,3 +66,39 @@ necessary. This may change in the future.
...
+
+   def _process_docstring(app, lines, transform, options):
+       if transform != 'foo':
+           return
+
+       comment = '\n.join(lines)
+       comment = foo_transform(comment)
+       lines[:] = comment.splitlines()[:]
+
+   # conf.py can be turned into an extension by adding setup() function
```
BrunoMSantos, line 139: Not sure this is helpful. Shouldn't we steer the user to use `extensions = ['hawkmoth', '<extension>']` instead or is that not how it's supposed to work?

hawkmoth/ext/transformations/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    # Note: None is a valid value for no transformation.
+    transform = transformations.get(tropt)
+    if transform is None:
+        return
+
+    comment = '\n'.join(lines)
+    comment = transform(comment)
+    lines[:] = comment.splitlines()[:]
+
+def setup(app):
```
BrunoMSantos, line 24: I'd consider adding deprecation notes. This might be new code, but it's an old feature we want to get rid of.

doc/extension.rst
BrunoMSantos: Indeed, `noindex` does solve the compilation issue (caching aside), but it's a hack. This is now a configuration variable of the new extension not of hawkmoth itself, which is the scope of this section.

Maybe we could replace this with a note informing that hawkmoth currently enables a built in extension by default to assist in transition and link it there so the user can find additional configuration options. It would be more accurate and avoid this oddity.

Note that the compiled docs (with links) are ok, but if you took the plain text or a printed format, the link below would still be a bit ambiguous.

hawkmoth/ext/javadoc/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2023, Jani Nikula <jani@nikula.org>
+# Licensed under the terms of BSD 2-Clause, see LICENSE for details.
+
+from hawkmoth.util.doccompat import javadoc_liberal
```
BrunoMSantos, line 4: What do you think of moving this code out of `hawkmoth.util` altogether?

doc/built-in-extensions.rst
jnikula: Right, maybe it's more confusing than helpful. The idea was to describe how people can convert their existing solution to an extension using the event as simply as possible.

hawkmoth/ext/javadoc/__init__.py
jnikula: Yes, this is what I want to do eventually, but didn't want to try to cram those changes here.

hawkmoth/ext/transformations/__init__.py
jnikula: I just get an uneasy feeling forcing a number of consecutive compat/transform changes on users...

doc/extension.rst
jnikula: > Maybe we could replace this with a note informing that hawkmoth currently enables a built in extension by default to assist in transition and link it there so the user can find additional configuration options.

Heh, well, the above was the goal of adding this here.

doc/built-in-extensions.rst
```diff
@@ -6,6 +6,27 @@ Built-In Extensions
...
+
+.. code-block:: python
+
+   extensions.append('hawkmoth.ext.javadoc')
+
+.. py:data:: hawkmoth_javadoc_transform
+   :type: list
+
+   List of transforms to convert, defaults to ``['javadoc']``. If the
+   ``transform`` option is not in in this list, do nothing.
```
BrunoMSantos, line 28: Hmm, this reads a bit funny and looks more complicated than it should. What would be the use case for the list?

`transform` is to be a string that needs to match the right trigger for this extension. Even if we make that trigger configurable (maybe a good idea as long as we support per-directive transformations), it should be a string and the text should read something like 'the name to use for this transformation'.

Same for the napoleon one.

hawkmoth/ext/javadoc/__init__.py
BrunoMSantos: Fair enough, up to you then.

doc/built-in-extensions.rst
BrunoMSantos: Yeah, I got that, but it sounds kind of hackish and without context. I'd vote it out :)

hawkmoth/ext/transformations/__init__.py
BrunoMSantos: Do we have users? :sweat_smile: 

In any case, the warnings should be gated on actual use of these functions, them being there doesn't mean we're about to remove them next week and we haven't released v1... We can delay the deprecation though, it was only a suggestion.

doc/extension.rst
BrunoMSantos: What do you think of, instead of:
```rst
.. py:data:: variable_that_does_not_really_belong_here_anymore
   :hack:
```
this:
```rst
.. note::

   For backwards compatibility, hawkmoth automatically includes the extention
   :extension:`something`, which itself exposes additional configuration options.
```

Possibly adding 'This automatic inclusion and the extension itself is planned to be removed later on.' depending on how you want to handle that.

BrunoMSantos:
So this was much simpler than I'd imagined. Love it! I have nothing but a few nitpicks.

doc/built-in-extensions.rst
jnikula: Yeah, so maybe this isn't fully thought out yet. The idea is, you've got documentation with `:transform: something-or-other` all over the place, and want to switch to using the built-in extension. But if it only accepts `javadoc` and nothing else, you're going to have to replace everything in flag day style.

Granted, looks like the upstream autodoc doesn't have anything like this. Even napoleon opportunistically converts if it can, and there's no further configuration, just adding `sphinx.ext.napoleon` to `extensions`.

BrunoMSantos: Registering the exact same transformation under multiple keys doesn't strike me as something we want to / should support in new APIs and I'd be surprised if anyone found / would find a reasonable use case for such a thing.

Still, there's already an extension providing backwards compatibility, why compromise the new stuff?

I'd say:
1. Keep supporting per directive, explicit transformations despite what napoleon does for instance. I think it's valuable and it's simple enough.
   The only thing here that makes me question this is that 3rd party plugins are not constrained to respect `:transform:` and I can't see a way of enforcing it without gating the emission of the event, which would be even worse I think.
3. Being a new API, keep it free from old stuff and as simple / elegant as possible. I'd argue the trigger for our built in extensions should _not_ be configurable at all until we find a strong reason for it and definitely not a list.

jnikula:
> So this was much simpler than I'd imagined. Love it! I have nothing but a few nitpicks.

I guess one question I have is the event interface. `hawkmoth-process-docstring(app, lines, transform, options)`. Contrast with https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-docstring

Is that good enough? Feels like the "what", "name", and "obj" parameters in `autodoc-process-docstring` would only be meaningful if we passed the cursor. But I'm not sure I want to go there yet. And that makes me wonder about extending later, and more backward incombatible changes... (do I worry too much?)

BrunoMSantos:
> > So this was much simpler than I'd imagined. Love it! I have nothing but a few nitpicks.
> 
> I guess one question I have is the event interface. `hawkmoth-process-docstring(app, lines, transform, options)`. Contrast with https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-docstring
> 
> Is that good enough? Feels like the "what", "name", and "obj" parameters in `autodoc-process-docstring` would only be meaningful if we passed the cursor. But I'm not sure I want to go there yet. And that makes me wonder about extending later, and more backward incompatible changes... (do I worry too much?)

Hmm, I'm generally ok with fast paced deprecation given the relative infancy of the project and small to non existent user base. But adding APIs that we already predict will become obsolete, that I don't like so much. And you're right, I hadn't considered that.

However, we can make this change as it is while explicitly advertising that the extension API is not stable and maybe adding some 'experimental' notice if anyone else other than us uses the API (if that's possible). As long as we maintain our built in extensions, there should be no regressions whatsoever within our stable API, but we gain a major clean up and the possibility to experiment with plugins internally even while no one else can trust the API. I'd say that's very acceptable.

It would mean no deprecation warnings on that extension though, despite my comment to that effect.

Oh, and side note, I'm exploring a solution for the cursor thing, before I tackle the C++ namespaces. Still early to know if it will pan out though.

jnikula:
Btw with the events, it's trivial to add new events and keep the old ones. So you could have the current event that is not aware about cursors, and then have another event, with a different name, that is aware of cursors. You could call them back to back, and there's no issue.

BrunoMSantos:
> Btw with the events, it's trivial to add new events and keep the old ones. So you could have the current event that is not aware about cursors, and then have another event, with a different name, that is aware of cursors. You could call them back to back, and there's no issue.

Even better, though it might be wise to safeguard ourselves this time anyway. Save us the trouble of supporting the old extension far into the future.

doc/extension.rst
jnikula: Resolved in #131 by just adding a reference to configuration in built-in extensions.

doc/extension.rst
jnikula: Resolved in #131 by just adding a reference to configuration in built-in extensions.

doc/built-in-extensions.rst
jnikula: Dropped in #131.

hawkmoth/ext/transformations/__init__.py
jnikula: Cleaned up the documentation text in #131, but opted not to deprecate this yet. The documentation implies this is going away in the future.

hawkmoth/ext/javadoc/__init__.py
jnikula: I've left this for future cleanup.

doc/built-in-extensions.rst
jnikula: Converted the transform names to strings in #131, and did the matching with simple `!=`. Clarified documentation as well.

I'm still keeping the simple configuration options though, I hope that's fine.

jnikula:
Closing, superseeded by #131.

PR #130, opened 2023-01-27
Clean up tests

BrunoMSantos:
Clean up attempt for the test/ folder. Basically tests are split in 3 categories for now: 'c', 'cpp' and 'examples', but nothing prevents us from breaking them down or nesting things further. Hopefully it will make it easier to find our bearings and reduce the amount of information encoded in file names, which was getting weird.


test/testenv.py
```diff
@@ -38,9 +38,10 @@ def __init__(self, filename):
         with open(filename, 'r') as f:
             self.options = strictyaml.load(f.read(), self._options_schema).data
         self.testid = os.path.splitext(os.path.basename(self.filename))[0]
+        self.namespace = os.path.relpath(os.path.dirname(self.filename), testdir)
 
     def get_testid(self):
-        return self.testid
+        return self.namespace + ': ' + self.testid
```
jnikula, line 45: My only problem with the series is that one space there `': '`. :)

If you want to run a single test with `pytest -k EXPRESSION` you can currently just copy-paste the name, e.g. `pytest -k test_directive_text[cpp-template]` and it works. However with the space I couldn't even figure out how to quote it for the expression. Just dropping the space makes it work.

I'm not sure I fully appreciate the definition of EXPRESSION at https://docs.pytest.org/en/latest/reference/reference.html#command-line-flags, and I'm not fussy about it as long as it's easy to copy-paste the id to run a single test case.

BrunoMSantos: I have to claim ignorance on this one, I just keep running all tests with make until they pass, quick enough that I haven't looked those alternatives up :sweat_smile: 

I wonder if we should instead use the path as the id. I.e. `path/to/file.ext` instead of `path/to: file`. What do you think?

jnikula: *shrug* 

Maybe drop the extension? `path/to/testcase`?

BrunoMSantos: I'll push a fix.

PR #131, opened 2023-01-29
Process docstring event v2

jnikula:
v2 of #129 hopefully with all the comments resolved.

I've added some commits cleaning up related documentation, and expanded on the built-in extension usage. There's more to do, surely, but I'm hoping this is a good first step.

I also pushed the rebased but otherwise unchanged version of #129 to `process-docstring-event-v1-rebased` which may be helpful in comparing the differences between the two.

### DIFF
--- a/doc/built-in-extensions.rst
+++ b/doc/built-in-extensions.rst
@@ -60,6 +60,58 @@ For example:
       :file: file.c
       :transform: javadoc
 
+.. _hawkmoth.ext.napoleon:
+
+hawkmoth.ext.napoleon
+---------------------
+
+This extension provides a bridge from Hawkmoth to the
+:external+sphinx:py:mod:`sphinx.ext.napoleon` extension, using the
+:event:`hawkmoth-process-docstring` event, to support Napoleon style
+documentation comments.
+
+Installation and configuration in ``conf.py``:
+
+.. code-block:: python
+
+   extensions.append('hawkmoth.ext.napoleon')
+
+.. py:data:: hawkmoth_napoleon_transform
+   :type: str
+
+   Name of the transformation to handle. Defaults to ``'napoleon'``. Only
+   convert the comment if the ``transform`` option matches this name, otherwise
+   do nothing.
+
+For example:
+
+.. code-block:: python
+   :caption: conf.py
+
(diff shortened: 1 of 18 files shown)


doc/built-in-extensions.rst
```diff
@@ -0,0 +1,70 @@
...
+This extension handles the :py:data:`cautodoc_transformations` feature, using
+the :event:`hawkmoth-process-docstring` event.
+
+.. note::
+
+   Going forward, it's recommended to handle transformations using the event
+   directly instead of :py:data:`cautodoc_transformations`. This built-in
+   extension provides backward compatibility for the functionality.
+
+   For now, this extension is loaded by default, and the below installation step
```
BrunoMSantos, line 129: The installation step below*

doc/extension.rst
```diff
@@ -31,7 +34,10 @@ For example:
 Configuration
 -------------
 
-The extension has a few configuration options that can be set in ``conf.py``:
+The extension has a few configuration options that can be set in ``conf.py``.
+
+See also additional configuration options in the :ref:`built-in extensions
+<built-in-extensions>`.
```
BrunoMSantos, line 40: I'm really nitpicking here, but this feels inferior to the admonition style example I made. Particularly, it's less direct in stating that there is a specific extension which is temporarily enabled by default and thus its configuration variables are a non-negotiable addition to these ones. I think it's that extension that we need a reference for here, not the whole set.

Than again... meh.

BrunoMSantos:
I was expecting some notice of unstable API for the event. After what we discussed I think it would be a good idea, even if it will be relatively easy to support these APIs well into the future. I'll trust your judgment on that one though.

Found a tiny grammatical and one other nitpick, but it's good really. Certainly no need for another round.

This project is starting to shape up :)

jnikula:
> I was expecting some notice of unstable API for the event. After what we discussed I think it would be a good idea, even if it will be relatively easy to support these APIs well into the future. I'll trust your judgment on that one though.
> 
> Found a tiny grammatical and one other nitpick, but it's good really. Certainly no need for another round.

I fixed the grammar, and merged this. We can tweak the rest before the next release if we want.

Thanks for the review!

> This project is starting to shape up :)

:+1:

PR #134, opened 2023-01-30
changelog: start listing changes since 0.12

jnikula:
Start listing changes since 0.12, with hopes of gearing towards a 0.13 release in the near future. There's been a lot of changes since 0.12, maybe more than in any other release, and I'd like to get it out sooner rather than later. (Release early, release often, and all that.)

@BrunoMSantos How does the hat-tip to your employer look like? 

I think the main blockers for 0.13 are documenting the C++ support, as well as adding any C++ features that you think we absolutely need before making a release. Not to put any pressure on you. ;)

I'm also fine with downgrading "Support for documenting C++" to "Initial support for documenting C++", depending on the level of support and/or boasting.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,39 @@ Unreleased `master`_
 
 :Date: YYYY-MM-DD
 
+Added
+~~~~~
+
+* Early support for documenting C++ (contributions courtesy of `Critical Software`_)
+* Support for extending documentation comment parsing and transformations via
+  ``hawkmoth-process-docstring`` event
+* ``hawkmoth_transform_default`` configuration option for the
+  ``hawkmoth-process-docstring`` event
+* ``hawkmoth_root`` configuration option to replace ``cautodoc_root``
+* ``hawkmoth_clang`` configuration option to replace ``cautodoc_clang``
+* Built-in extensions for Javadoc and Napoleon comment handling
+
+.. _Critical Software: https://www.criticalsoftware.com/
+
+Changed
+~~~~~~~
+
+* Typedefed anonymous struct, union, and enum parsing to be more explicit
+* ``cautodoc_transformations`` handling moved to a built-in extension
+* Lots of test suite refactoring and cleanups
+
+Deprecated
+~~~~~~~~~~
+
+* ``cautodoc_root`` configuration option in favour of ``hawkmoth_root``
+* ``cautodoc_clang`` configuration option in favour of ``hawkmoth_clang``
+
+Removed
+~~~~~~~
+
+* ``cautodoc_compat`` configuration option
+* ``compat`` directive option
+
 Hawkmoth `0.12.0`_
 ------------------
 


BrunoMSantos:
> Start listing changes since 0.12, with hopes of gearing towards a 0.13 release in the near future. There's been a lot of changes since 0.12, maybe more than in any other release, and I'd like to get it out sooner rather than later. (Release early, release often, and all that.)

Indeed!

> @BrunoMSantos How does the hat-tip to your employer look like?

Think it's good given what I discussed with them.

> I think the main blockers for 0.13 are documenting the C++ support, as well as adding any C++ features that you think we absolutely need before making a release. Not to put any pressure on you. ;)
> 
> I'm also fine with downgrading "Support for documenting C++" to "Initial support for documenting C++", depending on the level of support and/or boasting.

Right, the critical piece missing is namespaces I think, but I haven't made it a priority. I'm trying to improve a few things in the parser 1st. How do you feel about waiting until this next weekend? If it's not going fast enough, we release earlier, otherwise we may attempt a more complete release (C++ wise).

I'd also be ok with documenting what we have and leave a loud warning that it is experimental with a list of stuff we know is missing. Just the other day I remembered a few more missing things in fact, though nothing nearly as critical as namespaces.

CHANGELOG.rst
```diff
@@ -6,6 +6,39 @@ Unreleased `master`_
 
 :Date: YYYY-MM-DD
 
+Added
+~~~~~
+
+* Support for documenting C++ (contributions courtesy of `Critical Software`_)
```
BrunoMSantos, line 12: Not sure if you intend to merge this PR as it is for us to add things as we go along (not a bad idea by the way, and do so from now on for upcoming releases).

Anyway, if that's the case, I would only ask you to temper this sentence. Call it 'early support for' or 'cpp domain enablement' for now.

jnikula:
> Right, the critical piece missing is namespaces I think, but I haven't made it a priority. I'm trying to improve a few things in the parser 1st. How do you feel about waiting until this next weekend? If it's not going fast enough, we release earlier, otherwise we may attempt a more complete release (C++ wise).

I'm not *that* much in a rush, I'm leaning towards releasing something that's usable for documenting C++. Doesn't have to be complete, but I also don't want folks to try it out and think it's unusable. It's fine if people think "this doesn't cover the obscure feature I need, bummer" but less so if they go "can't do anything, what a waste of time to even try".

I'm also depending on you to define what "usable" means here. :)

CHANGELOG.rst
jnikula: I think I'll call it "early" or something for now, and we can adjust it later before the release, depending on what we have.

It would be great to update the changelog with the changes as they come instead of right before release, but nobody ever likes to write changelogs. :)

BrunoMSantos:
> > Right, the critical piece missing is namespaces I think, but I haven't made it a priority. I'm trying to improve a few things in the parser 1st. How do you feel about waiting until this next weekend? If it's not going fast enough, we release earlier, otherwise we may attempt a more complete release (C++ wise).
> 
> I'm not _that_ much in a rush, I'm leaning towards releasing something that's usable for documenting C++. Doesn't have to be complete, but I also don't want folks to try it out and think it's unusable. It's fine if people think "this doesn't cover the obscure feature I need, bummer" but less so if they go "can't do anything, what a waste of time to even try".
> 
> I'm also depending on you to define what "usable" means here. :)

Right, then I'd say we need namespaces for sure. It's an almost universal pattern to put everything in namespace blocks, and we don't recurse into those at the moment.

jnikula:
@BrunoMSantos Also, let me know if there's anything I could do to help. Like, maybe I could pick up what you had for documentation and rebase and polish that? Or any blockers?

jnikula:
> Not sure there's a huge benefit in doing this before or after the namespace thing, but I was nerd snipped into it and that's what I'm going with sweat_smile

I feel you! xD

BrunoMSantos:
> Thanks, I'll push the documentation part in case you want to have a look then. I don't think I changed it much since the version you saw.

Pushed here: https://github.com/BrunoMSantos/hawkmoth/tree/cpp-docs

Issue #92, opened 2022-10-27
change _Bool to more readable bool


jnikula:
Opening an issue for merge request #77. I want the feature, but I think the implementation is too trivial. I actually have some code parsing what's in the source to see if `_Bool` should be preserved, but hitting some issues.

Original comment:

Looks like the root cause is that `<stdbool.h>` defines `bool` instead of typedeffing it. Only the preprocessor sees `bool`, and from the compiler's perspective, it's `_Bool`. Contrast with `<stdint.h>` which typedefs e.g. `uint32_t`.

The trouble with your fix is that it's very specific to this particular problem. What if the source code had `_Bool`? Are there other cases where we should use whatever is in the source (and what the preprocessor sees) instead of the underlying type? I.e. should we look at the tokenized input instead?

_Originally posted by @jnikula in https://github.com/jnikula/hawkmoth/issues/77#issuecomment-1013650394_

BrunoMSantos:
With no prejudice towards your other points, I have an issue with the potential lack of consistency in this. If a user used a macro to specify a type conditional on (e.g.) architecture, it wouldn't have the same treatment as this `bool` one. Typedefs could solve it at the expense of code changes, but defines for function names resolving to compiler intrinsics for instance wouldn't have that luxury.

So maybe we need a way of extending this mechanism with user specified macros that should not be expanded either. I would probably use it in some places. We'd end up with symmetrical `_unexpand_(stdc|user)_macros` functions. What do you think?

jnikula:
Frankly, if the choice were between adding a generic user configurable mechanism or doing nothing, I think I'd choose to do nothing. I just consider `bool` vs. `_Bool` to be enough of a special case, and simple enough to implement.

Well, maybe. I got it almost working, but was hitting issues with struct members. :/

BrunoMSantos:
Yeah, the counter argument of complexity is a mighty one. I'll concede until we have concrete use cases at least.

jnikula:
So I have an updated version of #135 locally, with a new built-in extension to convert _Bool to bool, but I still think as a solution to this particular problem it's overkill. Maybe we want that anyway, as a generic thing, but perhaps changing _Bool to bool if bool was used in code is the sane default?

BrunoMSantos:
I think we need something indeed, but more in line with passing a cursor object to the event handler such that the event knows what is being expanded to what, if anything, and contextually make the right choice instead of a blind search & replace.

[...]

PR #135, opened 2023-01-30
Hawkmoth process signature proof-of-concept

jnikula:
I looked at #92 when going through the issues, and did a quick hack towards a proof-of-concept `hawkmoth-process-signature` event.

I'm not pursuing this for the next release, but it's something to consider in the future. Looks like this could be fairly simple. You can pretty nicely convert `_Bool` to `bool` if you like, with no hardcoded stuff in the parser. Maybe users want to annotate other stuff (where the clang cursor might be useful too).

Food for though, @BrunoMSantos

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -211,6 +211,12 @@ intersphinx_mapping = {
     'sphinx': ('https://www.sphinx-doc.org/en/master', None)
 }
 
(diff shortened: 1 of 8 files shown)


BrunoMSantos:
Yup, this would settle one of the points where we disagreed: extensibility. Best of both worlds :)

PR #137, opened 2023-02-04
doc: document the C++ support

jnikula:
Document the C++ support, and clean up and clarify documentation while at it.

I went through all the documentation and tweaked stuff all around the C++ changes. The first sentence in the `index.rst` was a mouthful even before the C++ stuff, so I simplified it a lot. On the other hand, I added more details and references here and there.

I opted to leave out any warnings about rudimentary C++ support for now. We can add that before the release depending on what we have.

### DIFF
--- a/doc/extension.rst
+++ b/doc/extension.rst
@@ -1,12 +1,13 @@
 .. _extension:
 
-C Autodoc Extension
-===================
+Autodoc Extension
+=================
 
 Hawkmoth provides a Sphinx extension that adds :ref:`new directives
-<directives>` to the Sphinx :external+sphinx:ref:`C domain <c-domain>` to
-incorporate formatted C source code comments into a document. Hawkmoth is Sphinx
-:py:mod:`sphinx.ext.autodoc` for C.
+<directives>` to the Sphinx :external+sphinx:ref:`C <c-domain>` and
+:external+sphinx:ref:`C++ <cpp-domain>` domains to incorporate formatted C and
+C++ source code comments into a document. Hawkmoth is Sphinx
+:py:mod:`sphinx.ext.autodoc` for C/C++.
 
 For this to work, the documentation comments must of course be written in
 correct reStructuredText. See :ref:`documentation comment syntax <syntax>` for
--- a/doc/directives.rst
+++ b/doc/directives.rst
@@ -4,18 +4,21 @@ Directives
 ==========
 
 Hawkmoth provides several new directives for incorporating :ref:`documentation
-comments <syntax>` from C source files into the reStructuredText document. There
-are three main types of directives, for incorporating documentation from entire
-files, for single objects, and for composite objects optionally with members.
+comments <syntax>` from C/C++ source files into the reStructuredText document.
+There are three main types of directives, for incorporating documentation from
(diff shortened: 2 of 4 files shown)


doc/extension.rst
```diff
@@ -67,7 +68,11 @@ See also additional configuration options in the :ref:`built-in extensions
 
    A list of arguments to pass to ``clang`` while parsing the source, typically
    to add directories to include file search path, or to define macros for
-   conditional compilation. No arguments are passed by default.
+   conditional compilation.
+
+   No arguments are passed by default when processing C domain
+   directives. ``-xc++`` is passed as the first argument when processing C++
+   domain directives.
```
BrunoMSantos, line 75: Not sure about this paragraph. It's technically accurate, but strictly from the view of a user, I worry it raises more questions than not.

BrunoMSantos:
Other than that minor comment, I think it's pretty clear.

> I opted to leave out any warnings about rudimentary C++ support for now. We can add that before the release depending on what we have.

I'd also argue against this one. You raised the good point before that merging this will expose these docs to some extent, even if there is no release yet. In fact I've just remembered that the official read the docs documentation is generated on the tip of master, we should probably fix that so that it happens only for tagged releases.

BrunoMSantos:
Other than that minor comment, I think it's pretty clear.

> I opted to leave out any warnings about rudimentary C++ support for now. We can add that before the release depending on what we have.

I'd also argue against this one. You raised the good point before that merging this will expose these docs to some extent, even if there is no release yet. In fact I've just remembered that the official read the docs documentation is generated on the tip of master, we should probably fix that so that it happens only for tagged releases.

jnikula:
> Other than that minor comment, I think it's pretty clear.
> 
> > I opted to leave out any warnings about rudimentary C++ support for now. We can add that before the release depending on what we have.
> 
> I'd also argue against this one. You raised the good point before that merging this will expose these docs to some extent, even if there is no release yet.

That was in reference to the README.rst changes, which I left out.

> In fact I've just remembered that the official read the docs documentation is generated on the tip of master, we should probably fix that so that it happens only for tagged releases.

So it generates both "latest" and "stable" documentation, with the former being the tip of master, and the latter being the latest tagged release. As a quick fix, I changed the redirect of https://hawkmoth.readthedocs.io/ from latest to stable.

"latest" was a poor choice of wording in that it doesn't really convey "unreleased" like it should. (I don't even remember if that was my wording or RTD default or what.)

With the latest being deployed on github pages at https://jnikula.github.io/hawkmoth/ (though not referenced anywhere yet) it might be prudent to sunset latest from RTD.

doc/extension.rst
jnikula: I left this change out.

jnikula:
I guess we both ran out of steam/time a bit with this...

I made the minor update about clang options. With the version now more prominently displayed in the documentation sidebar, and the default being "stable" at readthedocs, I think it should be good enough now, to make some forward progress.

BrunoMSantos:
> I guess we both ran out of steam/time a bit with this...

Yeah, RL hasn't helped. I've kept an eye on this, but not in long enough stints to see anything through...

Nice to see the documentation improve though! Looks good to me.

PR #138, opened 2023-02-04
c:autotext and cpp:autotext directives for top level comments

jnikula:
Add support for picking up individual top level comments not related to any cursors or symbols. This was neglected when the cursor-specific auto* directives were added. The top level comments could only be extracted using `c:autodoc`.

The implementation is incredibly simple, and fits right in the existing framework.

The only snag is the requirement to figure out names for the documentation comments to be able to reference them, and this means parsing the comments somehow. This is something that I generally want to avoid like the plague, but perhaps the proposal here is simple enough. This doesn't change the parser or the identification of top level comments, and only the top level comments are "parsed" for names when needed, and it's a simple regex.

What do you think?

### DIFF
--- a/hawkmoth/docstring.py
+++ b/hawkmoth/docstring.py
@@ -198,6 +198,35 @@ class TextDocstring(Docstring):
     _indent = 0
     _fmt = '\n'
 
+    def get_name(self):
+        """Figure out a name for the text comment based on the comment contents.
+
+        The name is the sub-string starting from the first alphanumeric
+        character in the comment to the next :, ., or newline.
+
+        This sensibly covers cases like reStructuredText hyperlink targets::
+
+            .. _Foo Bar:
+
+        and section titles::
+
+            Foo Bar
+            =======
+
+        and just first sentences::
+
+            Foo Bar. Blah.
+
+        Not perfect, but good enough.
+        """
+        # If the parser passed in a name, use it (unlikely)
+        if self._name:
+            return self._name
+
+        mo = re.search(r'[\W_]*(?P<name>\w[^:.\n\r]*)', self._text)
+
+        return mo.group('name') if mo else None
+
 class VarDocstring(Docstring):
     _indent = 1
     _fmt = '\n.. {domain}:var:: {ttype}{type_spacer}{name}\n\n'
--- a/doc/directives.rst
+++ b/doc/directives.rst
@@ -172,3 +172,41 @@ The directives support all the same directive options as :rst:dir:`c:autodoc`,
       .. c:autoenum:: example_enum
          :file: example_file.c
          :members: CONSTANT_ONE, CONSTANT_TWO
+
+Text Documentation Comments
+---------------------------
+
+.. rst:directive:: .. c:autotext:: name
+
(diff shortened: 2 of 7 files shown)


hawkmoth/docstring.py
```diff
@@ -198,6 +198,35 @@ class TextDocstring(Docstring):
...
+        and section titles::
+
+            Foo Bar
+            =======
+
+        and just first sentences::
+
+            Foo Bar. Blah.
+
+        Not perfect, but good enough.
```
BrunoMSantos, line 220: Hmmm, this looks simple, but suddenly a user documents two files with the same initial line and Sphinx starts complaining about duplicate reference tags.

Food for thought: anonymous symbols get a completely unguessable and non permanent tag right now. Here, a filename might be a good alternative, but for anonymous types we'd need file and line number, which is maybe too brittle and still not 'permanent'.

A common solution to these cases might be to generate throw away, unguessable and non permanent tag by default if we must or none if we don't, and instruct the user to specify a tag in those cases as needed. E.g.:

```c
/**
 * .. _tag:
 * Top level comment.
 */
```

In the anonymous types' case this would produce a link to the body of the block rather than the declaration line though I think.

BrunoMSantos:
> What do you think?

You're right, we overlooked this one. But I don't think that strategy for references will work :/

hawkmoth/docstring.py
jnikula: > Hmmm, this looks simple, but suddenly a user documents two files with the same initial line and Sphinx starts complaining about duplicate reference tags.

That's actually not true. It's not automatically generating any reference targets for Sphinx.

The user can add hyperlink targets themselves, and if they add it as the first thing in the comment, it also works for the purpose of selecting the comment. But then they're adding the targets, and it's their responsibility to handle any conflicts.

BrunoMSantos: Bah, sorry, this case has no directive of course, so there isn't an automatic reference tag being generated.

As for filtering the right comments, right now the file argument is mandatory, but even afterwards it can be used to disambiguate if need be. That still leaves comments in the same file starting the same way as a potential issue, right? Not sure we want to solve that though.

jnikula: > That still leaves comments in the same file starting the same way as a potential issue, right? Not sure we want to solve that though.

Even that's only a problem if the user wants to pick one or the other; I think with the current implementation it'll probably pick both.

I'm in no way claiming this naming system is ideal, but AFAICS everything else pretty much requires proper parsing of the comment for some structured identifier for the comment. For example requiring some `:meta id: name` field list and parsing using `extract_metadata()` from `sphinx.util.docstrings`. But that also requires stripping the comment indicators first and overall complicates matters.

I think a generic documentation comment would typically have a heading or a descriptive first sentence or a reference target, and that just could work as a name.

BrunoMSantos: Yup, I see your points. But if you don't mind, let it marinate a bit longer. This will be part of our API... Better think it through now than later.

jnikula:
Superseeded by #161

PR #139, opened 2023-02-05
Doc alabaster theme

jnikula:
I was reminded by https://github.com/jnikula/hawkmoth/pull/137#issuecomment-1416878236 that I've got this branch where I'm experimenting with displaying documentation version more prominently and potentially adding multiple version support. I'm not quite there yet, but here's some early stuff. Not much point in hacking on the alabaster theme to display versions if we disagree on switching to it!

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -127,9 +127,15 @@ html_theme_options = {
 # defined by theme itself.  Builtin themes are using these templates by
 # default: ``['localtoc.html', 'relations.html', 'sourcelink.html',
 # 'searchbox.html']``.
-#
-# html_sidebars = {}
-
+html_sidebars = {
+    '**': [
+        'about.html',
+        'navigation.html',
+        'relations.html',
+        'searchbox.html',
+        'donate.html',
+    ]
+}
 
 # -- Options for HTMLHelp output ---------------------------------------------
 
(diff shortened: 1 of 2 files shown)


BrunoMSantos:
The changes look good to me, though I haven't delved into the theme configuration options to know any better.

> Not much point in hacking on the alabaster theme to display versions if we disagree on switching to it!

It's ok with me, but I'm curious, why are you making this a requirement to get the version stuff working?

jnikula:
> The changes look good to me, though I haven't delved into the theme configuration options to know any better.
> 
> > Not much point in hacking on the alabaster theme to display versions if we disagree on switching to it!
> 
> It's ok with me, but I'm curious, why are you making this a requirement to get the version stuff working?

The multiversion thing is a can of worms, I'm afraid. I'm not really happy with anything that's readily available, and the RTD flyout menu is their proprietary thing. So I'm looking at hacking the theme to add the version menu to it. It's not necessarily tied to the alabaster theme, but it's easier to work from with just one theme in mind. Alabaster is strictly not part of Sphinx, but it's the default theme and always ships with Sphinx, so dropping the sphinx_rtd_theme reduces moving parts.

jnikula:
> The changes look good to me

I merged with that, thanks, and now you can compare the differences at RTD:

- https://hawkmoth.readthedocs.io/en/stable/
- https://hawkmoth.readthedocs.io/en/latest/

jnikula:
Also posted #140 to demo how it would look like.

PR #140, opened 2023-02-05
doc: add documentation version to sidebar

jnikula:
Bump version to 0.13.0.dev0, indicating developmental release per [1], and add the version to the theme sidebar.

[1] https://peps.python.org/pep-0440/#developmental-releases

The idea is to expand the Version sidebar to Versions, and add multiple versions there.


BrunoMSantos:
This looks clean enough. Feel free to merge once you're confident this approach works for the multiple versions.

PR #142, opened 2023-03-19
Doc cpp

jnikula:
Add minimal C++ class example, update README and briefly describe the limitations of C++ support, in preparation for making a release with whatever we have now.


doc/examples.rst
```diff
@@ -237,6 +237,33 @@ Output
...
 
 
+Class
+-----
+
+Source
+~~~~~~
+
+.. literalinclude:: ../test/examples/class.cpp
+   :language: C
```
BrunoMSantos, line 247: I'm pretty sure pigments supports C++ syntax or is this how that is specified?

doc/index.rst
```diff
@@ -10,6 +10,12 @@ documentation, respectively. In short, Hawkmoth is Sphinx Autodoc for C/C++.
 Hawkmoth aims to be a compelling alternative for documenting C and C++ projects
 using Sphinx, mainly through its simplicity of design, implementation and use.
 
+.. note::
+
+   The C++ support is still in early stages of development, and lacks some
+   fundamental features such as handling namespaces and documenting C++ specific
+   features other than classes.
```
BrunoMSantos, line 17: Not sure the part after 'and' is quite clear or accurate, though I don't think it needs to be in this case as it really is just a waiver.

For the record though, and while I haven't looked at this in some time, these are all the things I had on my todo list and at the time I felt confident there wouldn't be much more aside from the newest standards perhaps:

Important:
* namespace traversal
* using keyword

Not important:
* lambdas are treated as variables, not functions, though I'm not sure we want to go down that rabbit hole anyway, so much so that one could say the same about `std::function` objects or any user implementation of a similar thing.
* `friend` keyword, though it's an eye sore and no one should use it :shrug:

jnikula: > Not sure the part after 'and' is quite clear or accurate, though I don't think it needs to be in this case as it really is just a waiver.

Good enough to merge?

jnikula:
> Hate to admit it but I really did loose steam on this. I take it you intend to issue a new release in the current state? If so, I think it's a good idea too and the changes look fine to me despite the couple of comments.

I think we had a good sprint there, but both ran out of steam. And I think it's fine. Don't worry about it.

But I also don't want to drag another release any longer. There's a bunch of good stuff that could use a release. So let's document what we have, and push it out.

doc/examples.rst
BrunoMSantos: Ah, of course, forgot this was generated. Sure then.

doc/index.rst
BrunoMSantos: Yeah, it's enough to scare most people away except those wanting to engage ;)

PR #143, opened 2023-03-22
update-examples: choose literalinclude language based on domain

jnikula:
Use proper language for literalinclude based on the domain. Warn about the same file being used for multiple domains, although C++ should handle syntax highlighting for both just fine.


BrunoMSantos:
Looks good to me ;)

PR #146, opened 2023-03-30
doc: require hawkmoth for the documentation build

jnikula:
Now that we have documentation deployed at GitHub Pages, and we no longer have to rely on Clang being available at Read The Docs in the future, require Hawkmoth for the documentation build.

The fallback automock has not been updated in ages, and the documentation build doesn't even work properly with that anymore. Just throw it out.


BrunoMSantos:
Yup, looks good to me too ;)

PR #147, opened 2023-03-30
Fix packaging

jnikula:
Ugh. We'll need a new release ASAP.


BrunoMSantos:
Didn't test it, but looks good. I also need to update the arch packages and start using the newer versions :/

PR #149, opened 2023-04-01
switch to src/ hierarchy

jnikula:
Trying to avoid the embarrassment of releasing completely broken packages.

With this, and 9ad923e5cf0f ("build: actually package hawkmoth.ext.*") reverted, `make test` actually fails as expected.

You may need to rm -rf the old venv first.


jnikula:
Looks like this needs to be split up to two separate pull requests, with the last commit in a separate pull request.

Makefile.local
```diff
@@ -38,6 +38,6 @@ test-upload:
 upload:
 	twine upload dist/*
 
-.PHONY: dist-tools dist test-upload upload
+.PHONY: build-tools build test-upload upload
```
BrunoMSantos, line 41: I usually follow a pattern like:

```make
.PHONY: target1
target1: ...
        ...

.PHONY: target2
target2: ...
        ...
```

Up to you, but it makes it much less likely these ever go out of sync again.

.github/workflows/makefile.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
       - name: Install dependencies (pip)
-        run: pip install --requirement requirements.txt --upgrade --upgrade-strategy eager
```
BrunoMSantos, line 29: Looking at the CI errors, my 1st suspicion would be that each CI step runs in a new shell and therefore the venv is not active for the `make test` step.

BrunoMSantos:
Looks good to me aside not passing CI. Also great that you found the root cause for the breakage in the tests.

Makefile.local
jnikula: Fixed that in #150

.github/workflows/makefile.yml
jnikula: Fixed that in #150 by using:
```
      - name: Test
        run: |
          . venv
          make test
```

PR #150, opened 2023-04-01
Src hierarchy part 1

jnikula:
Part 1 of #149 to hopefully pass CI.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,10 +22,10 @@ jobs:
           ref: ${{ matrix.ref }}
       - name: Install dependencies (apt)
         run: sudo apt-get install -y python3-clang python3-pip
-      - name: Install dependencies (pip)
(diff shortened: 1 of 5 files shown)


jnikula:
Fail, back to the drawing board.

jnikula:
@BrunoMSantos Seems to work now

BrunoMSantos:
Looks good! Not sure why it needs to be in 2 parts, but works for me.

jnikula:
> Looks good! Not sure why it needs to be in 2 parts, but works for me.

Yeah, maybe it wouldn't have needed to be, I had one idea why it failed and stuck to it even after reading your comment about the separate shells. Oh well.

BrunoMSantos:
> > Looks good! Not sure why it needs to be in 2 parts, but works for me.
> 
> Yeah, maybe it wouldn't have needed to be, I had one idea why it failed and stuck to it even after reading your comment about the separate shells. Oh well.

Yup, thought so, but it won't matter after the merge ;)

PR #151, opened 2023-04-01
move source under src/ part 2

jnikula:
The v0.13 release does not work at all, because the packaging failed to include all the new hawkmoth.ext sub-packages. The testing did not catch this, because we were testing from the source directly.

Switch to a src/ layout to force testing on the editable install added previously. Rationale [1][2][3].

[1] https://hynek.me/articles/testing-packaging/
[2] https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure
[3] https://docs.pytest.org/en/latest/explanation/goodpractices.html


jnikula:
Perhaps the `venv` script is a bit clunky, and long term we should probably use something like tox, but didn't find the time or energy to learn to use it now.

BrunoMSantos:
Good to go!

> Perhaps the `venv` script is a bit clunky, and long term we should probably use something like tox, but didn't find the time or energy to learn to use it now.

Functionally, it should be equivalent to passing different `requirements.txt` files to the venv scrip. And I kind of like knowing what's happening under the hood provided we're not reinventing any big wheels.

Then again, maybe the clean-up will be worthwhile if / when we test multiple configurations or maybe it provides some performance improvements relative to our simple script. Either way, I don't think we would be reaping a huge benefit from it for a single venv.

jnikula:
> Good to go!

Thanks again for the swift reviews! I'll probably do a v0.14.0 release soon.

> Functionally, it should be equivalent to passing different `requirements.txt` files to the venv scrip. And I kind of like knowing what's happening under the hood provided we're not reinventing any big wheels.

Yeah, it's not too many lines as it is.

> Then again, maybe the clean-up will be worthwhile if / when we test multiple configurations or maybe it provides some performance improvements relative to our simple script. Either way, I don't think we would be reaping a huge benefit from it for a single venv.

I actually quickly tried to use tox, and I got it sort of working, but I could not reproduce the packaging issue, like I could with plain `pip install --config-settings editable_mode=strict --editable .`. So I also could not see the benefit until it actually detects the issue we had!

PR #152, opened 2023-04-03
docker: fix make docker-test and docker-test-html targets

jnikula:
Switching to src/ hierarchy broke the docker-test and docker-test-html targets, as the old docker setup specifically used the hawkmoth package from the source, using a mount, and the main point of the src/ hierarchy was to make this difficult.

Instead, archive the source and expand it in the container, using the same editable build method. It's not perfect, but it works.

### DIFF
--- a/docker/Makefile.local
+++ b/docker/Makefile.local
@@ -1,29 +1,35 @@
 # -*- makefile -*-
-# Copyright (c) 2021, Jani Nikula <jani@nikula.org>
+# Copyright (c) 2021-2023, Jani Nikula <jani@nikula.org>
 # Licensed under the terms of BSD 2-Clause, see LICENSE for details.
 
 docker_dir := docker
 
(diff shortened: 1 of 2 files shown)


BrunoMSantos:
I confess I've used this maybe once to see that it worked and never again. If I remember correctly this was added due to some distros packaging an old version of Clang. Am I remembering it wrong? I assume it's still needed.

Anyway, code wise, looks good to me :+1:

jnikula:
> I confess I've used this maybe once to see that it worked and never again. If I remember correctly this was added due to some distros packaging an old version of Clang. Am I remembering it wrong? I assume it's still needed.

I don't really use it, but doesn't seem too much trouble to keep it working, in case a need arises.

The other docker images that I upload to docker hub I've pretty much decided to discontinue. Doesn't look like anyone's really using them. I think I'll just stop and see if someone ever complains.

> Anyway, code wise, looks good to me +1

Thanks, merged!

PR #153, opened 2023-04-21
test: add our own src/ directory to the search path

BrunoMSantos:
So I _finally_ got around to try and package the newest version for Arch and discovered I couldn't run the tests with a simple `make test`.

No idea just yet if this will break the CI. Not sure why / how it's running in CI either, though I confess I haven't looked too hard into it. Either way, I am positive that `make test` was trying to run my old system installation of hawkmoth and this fixes it for me.

Not so related, and not sure if you had already commented on that, but `sphinx-testing` is now properly deprecated. I'm currently running an unmaintained version. We're probably on borrowed time here and we'll have to convert to the built in `sphinx.testing` soon :/

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -2,6 +2,11 @@
 # Licensed under the terms of BSD 2-Clause, see LICENSE for details.
 
 import pytest
+import os
+import sys
+
+# Use 'our' version of the module.
+sys.path.insert(0, os.path.abspath('src'))
 
 # Enable introspection on assertions in testenv
 pytest.register_assert_rewrite('test.testenv')


BrunoMSantos:
Hmm, I thought that was meant to be covered by CI but that we would still support an easy and quick way of testing without it locally. Though I guess this would have to be done differently even then.

To tell the truth, I stay away from venvs as much as possible. But I guess it's easy enough to carry around a workaround on my side for now so I'll do that.

> Meh. I've looked at it several times, and it's not a simple conversion.

Yup, got the same impression :/

PR #156, opened 2023-05-24
RFC: Cache filename

jnikula:
Only tested locally, no documentation, but thoughts?

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -161,6 +161,12 @@ class _AutoSymbolDirective(_AutoBaseDirective):
     def _get_filenames(self):
         filename = self.options.get('file')
 
+        # Cache filename across directives in rst document
+        if filename:
+            self.env.temp_data['hawkmoth_current_file'] = filename
+        else:
+            filename = self.env.temp_data.get('hawkmoth_current_file')
+
         # Note: For the time being the file option is mandatory (sic).
         if not filename:
             self.logger.warning(':file: option missing.',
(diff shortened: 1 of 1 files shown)


src/hawkmoth/__init__.py
```diff
@@ -161,6 +161,12 @@ class _AutoSymbolDirective(_AutoBaseDirective):
     def _get_filenames(self):
         filename = self.options.get('file')
 
+        # Cache filename across directives in rst document
+        if filename:
+            self.env.temp_data['hawkmoth_current_file'] = filename
+        else:
+            filename = self.env.temp_data.get('hawkmoth_current_file')
```
BrunoMSantos, line 168: Right, so having looked into this I stand by my comment in #157: I don't like the idea of stateful directives too much, but I've also reconsidered things a bit. I would still personally vote this particular implementation down in favour of having a specialized directive to select a file.

Not only that would be much more explicit, we could later combine it with full lookup / cache as a way of speeding things up by limiting the search to a (set of) directory --- i.e.: `.. autoresolve:: some/dir`. And in that case, adding the new directive with support for a single file only could be done without implying future API changes.

Thoughts?

jnikula: > Right, so having looked into this I stand by my comment in https://github.com/jnikula/hawkmoth/issues/157: I don't like the idea of stateful directives too much. But I've also reconsidered things a bit, even if I would still personally vote this particular implementation down in favour of having a specialized directive to select a file.

Agreed. If there are going to be stateful directives, they should be explicit rather than implicit like this.

jnikula:
Superseded by #168

PR #158, opened 2023-06-30
build: add distclean target

jnikula:
'make clean' removes the editable install. Separate the removal of the build artefacts to 'make distclean'.


BrunoMSantos:
Fine by me!

PR #161, opened 2023-07-01
Autotext v2

jnikula:
I'm still pursuing #138. This improves and expands documentation and examples, and emphasizes that the "reference names" do not automatically become hyperlink targets.

### DIFF
--- a/doc/directives.rst
+++ b/doc/directives.rst
@@ -203,3 +203,50 @@ The directives support all the same directive options as :rst:dir:`c:autodoc`,
       .. c:autoenum:: example_enum
          :file: example_file.c
          :members: CONSTANT_ONE, CONSTANT_TWO
+
+Generic Documentation Comments
+------------------------------
+
+.. rst:directive:: .. c:autotext:: name
+.. rst:directive:: .. cpp:autotext:: name
+
+   Incorporate the text documentation comment identified by ``name`` in the file
+   ``file``. The ``file`` option is as in :rst:dir:`c:autovar`.
+
+   The ``name`` is derived from the first sentence of the comment, and may
+   contain whitespace. It starts from the first alphanumeric character,
+   inclusive, and extends to the next ``:``, ``.``, or newline, non-inclusive.
+
+   For example:
+
+   .. code-block:: c
+
+      /**
+       * This is the reference. This is not. It all becomes
+       * the documentation comment.
+       */
+
+   .. code-block:: rst
+
+      .. c:autotext:: This is the reference
+	 :file: example_file.c
+
+   Note that the above does not automatically create hyperlink targets that you
+   could reference from reStructuredText. However, reStructuredText hyperlink
+   targets work nicely as the reference name for the directive:
+
+   .. code-block:: c
+
+      /**
+       * .. _This is the reference:
+       *
+       * The actual documentation comment.
+       *
+       * You can use :ref:`This is the reference` to reference
+       * this comment in reStructuredText.
+       */
+
+   .. code-block:: rst
+
+      .. c:autotext:: This is the reference
+	 :file: example_file.c
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -312,6 +312,31 @@ Output
    :members:
 
 
+Generic Documentation Comments
+------------------------------
+
+Source
+~~~~~~
+
+.. literalinclude:: ../test/examples/autotext.c
+   :language: C
+   :caption: autotext.c
+
+Directive
+~~~~~~~~~
+
+.. code-block:: rest
+
+   .. c:autotext:: Hyperlink Target
+      :file: autotext.c
+
+Output
+~~~~~~
+
+.. c:autotext:: Hyperlink Target
+   :file: autotext.c
+
+
 Preprocessor
 ------------
 
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -191,6 +191,12 @@ class _AutoCompoundDirective(_AutoSymbolDirective):
 class CAutoDocDirective(_AutoDocDirective):
     _domain = 'c'
 
+class CAutoTextDirective(_AutoSymbolDirective):
+    # Allow spaces in the directive argument (the name)
+    final_argument_whitespace = True
+    _domain = 'c'
+    _docstring_types = [docstring.TextDocstring]
+
 class CAutoVarDirective(_AutoSymbolDirective):
     _domain = 'c'
     _docstring_types = [docstring.VarDocstring]
@@ -221,6 +227,12 @@ class CAutoEnumDirective(_AutoCompoundDirective):
 
 class CppAutoDocDirective(_AutoDocDirective):
     _domain = 'cpp'
+
+class CppAutoTextDirective(_AutoSymbolDirective):
+    # Allow spaces in the directive argument (the name)
+    final_argument_whitespace = True
+    _domain = 'cpp'
+    _docstring_types = [docstring.TextDocstring]
 
 class CppAutoVarDirective(_AutoSymbolDirective):
     _domain = 'cpp'
@@ -283,6 +295,7 @@ def setup(app):
     app.add_config_value('hawkmoth_transform_default', None, 'env', [str])
 
     app.add_directive_to_domain('c', 'autodoc', CAutoDocDirective)
+    app.add_directive_to_domain('c', 'autotext', CAutoTextDirective)
     app.add_directive_to_domain('c', 'autovar', CAutoVarDirective)
     app.add_directive_to_domain('c', 'autotype', CAutoTypeDirective)
     app.add_directive_to_domain('c', 'autostruct', CAutoStructDirective)
@@ -292,6 +305,7 @@ def setup(app):
     app.add_directive_to_domain('c', 'autofunction', CAutoFunctionDirective)
 
     app.add_directive_to_domain('cpp', 'autodoc', CppAutoDocDirective)
+    app.add_directive_to_domain('cpp', 'autotext', CppAutoTextDirective)
     app.add_directive_to_domain('cpp', 'autovar', CppAutoVarDirective)
     app.add_directive_to_domain('cpp', 'autotype', CppAutoTypeDirective)
     app.add_directive_to_domain('cpp', 'autostruct', CppAutoStructDirective)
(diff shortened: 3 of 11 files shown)


BrunoMSantos:
So, on this one, I see no big issue with implementation apart from the CI issue, which I only looked into briefly, but it seems you're already working on it.

But I keep thinking it's not nicely symmetrical with the Python autodoc and I'd like to throw a few ideas around, see what you think.

-->8--

So in Python's autodoc, there's `automodule` which only captures the 1st 'top level' docstring, other ones are impossible to extract as far as I can remember. That's fine for Python as that would not comply with its documentation guidelines. I do _not_ think we want to be so restrictive in C, but I mention it for context.

Python's `automodule` only exposes the module ('top level') docstring by default unless the `:members:` option is used.

I would like to see our `autodoc` be more like `automodule` in that it omits members by default, but that still leaves us to deal with multiple top level comments. And, presuming that there is an actual use case for selecting them individually, that's where I lack any better idea and where `autodoc` might start being a bit overloaded if we extend it any further.

Could there be a clean enough way of using `autodoc` for this purpose too somehow, without creating a new directive? If so, I would really have 0 reservations about it.

If not, then I can live with this as an alternative too.

jnikula:
> So, on this one, I see no big issue with implementation apart from the CI issue, which I only looked into briefly, but it seems you're already working on it.

This was a silly one, conflicting `example-priority: 80` and it lead to different sort order between them on different machines.

jnikula:
I think there's a use case for selecting multiple generic documentation comments individually. For example, you have different "sets" of functions in a file, and you want to describe those sets at the high level using generic comments. So agreed, we shouldn't restrict to just one.

I think it would be *possible* to reuse `c:autodoc` for this by introducing the `:members:` option. For structs etc. it's a comma-separated list, so could be one here too:

```
.. c:autodoc: file.c
   :members: a generic comment, another
```

However, adding `:members:` either makes it inconsistent with `automodule` when no `:members:` is included, or we have to break compatibility. I don't like either.

Another option is to introduce `c:automodule` with semantics closer to the Python domain directive, especially wrt `:members:`. But are generic comments "members"...?

Additionally, I'm working on making the `:file:` option actually optional for all the other directives. With that, including generic documentation comments becomes the odd one out if you need to use `c:autodoc` (or `c:automodule`) with the filename as argument for them. With a completely new directive such as `c:autotext` (and I'm open to other names!) we could drop the `:file:` the same way as for other directives.

BrunoMSantos:
> However, adding `:members:` either makes it inconsistent with `automodule` when no `:members:` is included, or we have to break compatibility. I don't like either.

On that specific topic, I would actually like to break compatibility as I'm sure I have mentioned before. I do think the behaviour should change when no `:members:` is present for better consistency.

> Another option is to introduce `c:automodule` with semantics closer to the Python domain directive, especially wrt `:members:`.

Note we did consider and I recall you were in agreement with renaming autodoc to automodule, but I eventually backed down from that or put it on hold at least as C++20 has 'modules' (which I don't know much about yet) which could lead to possible confusions.

> But are generic comments "members"...?

Yeah, that's the big question. Though you raise another interesting point further down.

For the record, when I referred to presuming the existence of a use case, I meant specifically the cherry picking of top level comments and not whether there could be more than one. I personally would be very happy with no cherry picking at all. E.g.:

```c
/** Foo. */

/** Fooer.*/
void fooer(void);

/** Bar. */

/** Barer.*/
void barer(void);
```

With `.. autodoc:: file.c` (no `:members:`), would always yield:

```rst
Foo.

Bar.
```

And although I admit others may see it differently, I also have a not so slight suspicion that such a desire would stem from wanting to shoehorn hawkmoth into a Doxygen environment as a somewhat easier way around the `@ingroup`. Not that it removes the urgency of having such a feature, but it does make me a bit sadder :laughing:

> Additionally, I'm working on making the `:file:` option actually optional for all the other directives. With that, including generic documentation comments becomes the odd one out if you need to use `c:autodoc` (or `c:automodule`) with the filename as argument for them. With a completely new directive such as `c:autotext` (and I'm open to other names!) we could drop the `:file:` the same way as for other directives.

That's true, unless `autodoc` would somehow be plastic enough to take a file or a comment ref, which was in part what I mentioned by excessive overloading of the directive.

-->8--

Anyway, I think I exhausted all I can say on the matter. I'm willing to concede that the best cherry picking option might well be this one. As for the directive name... dunno. Can't say I like having two very generic autodoc and autotext side by side. I also have no strong suggestions though.

Maybe `autoblock`? Particularly if we could implement a `:members:` option that would capture anything below it _until_ the next top level comment. That might even be a better `@ingroup` look alike actually. Haven't thought it through.

jnikula:
At the moment my order of preference is:

1. add dedicated `c:autotext` (or similar) directives
2. add `c:automodule` with `:members:` support
3. amend `c:autodoc` with `:members:`

jnikula:
> For the record, when I referred to presuming the existence of a use case, I meant specifically the cherry picking of top level comments and not whether there could be more than one. I personally would be very happy with no cherry picking at all. E.g.:
> 
> ```c
> /** Foo. */
> 
> /** Fooer.*/
> void fooer(void);
> 
> /** Bar. */
> 
> /** Barer.*/
> void barer(void);
> ```
> 
> With `.. autodoc:: file.c` (no `:members:`), would always yield:
> 
> ```
> Foo.
> 
> Bar.
> ```
> 
> And although I admit others may see it differently, I also have a not so slight suspicion that such a desire would stem from wanting to shoehorn hawkmoth into a Doxygen environment as a somewhat easier way around the `@ingroup`. Not that it removes the urgency of having such a feature, but it does make me a bit sadder 😆

I was going to switch off the computer by now, but one more reply. Just one, I promise! :)

I believe the order of stuff you have in a C source file is rarely the order in which you want to document them. That's possibly the biggest limitation in `c:autodoc`, even more so than not having fine grained control over the contents. And if you have someone willing to write documentation, you shouldn't force them to try to reorder the C source to make it nice. This is why I also think the generic documentation comments are unlikely to be placed in a way that is automatically natural for the documentation.

So with `c:autodoc` or `c:automodule`, you might also need to add them multiple times to cherry-pick the right comments in the right places in the documentation, and that doesn't really feel nice either.

To be continued... :)

BrunoMSantos:
> I believe the order of stuff you have in a C source file is rarely the order in which you want to document them.

Eheh spicy! I agree, and that would fall into the shoehorning category in my book :stuck_out_tongue:

Anyway, I get that we are pre-empting the need for adaptations to existing code bases as much as fixing an asymmetry in the current API. However unnecessary it might seem to me personally.

I would be interested in what you think of the autoblock idea. Other than that, I agree option 1 is the best form of cherry picking on the table right now.

jnikula:
> > I believe the order of stuff you have in a C source file is rarely the order in which you want to document them.
> 
> Eheh spicy! I agree, and that would fall into the shoehorning category in my book 😛

Heh, it was not intended to be spicy, just an observation of reality. Including people having strong opinions about both how to organize C source, and how to document it, and them conflicting badly! :smile: 

> 
> Anyway, I get that we are pre-empting the need for adaptations to existing code bases as much as fixing an asymmetry in the current API. However unnecessary it might seem to me personally.
> 
> I would be interested in what you think of the autoblock idea. Other than that, I agree option 1 is the best form of cherry picking on the table right now.

The downside with "block" is that my first thought was it was a verb "to block" something. Autotext, autosection, autochapter, ...?

BrunoMSantos:
> The downside with "block" is that my first thought was it was a verb "to block" something. Autotext, autosection, autochapter, ...?

I meant the idea of having a `:members:` option and including everything until but not including the next top level comment in its scope.

As for the name, in that scenario, 'text' would not the best option and 'block' (as a name) was the 1st thing to come to mind, but 'section' or 'chapter' sound equally decent to me too.

jnikula:
> I meant the idea of having a `:members:` option and including everything until but not including the next top level comment in its scope.

Whoops, I think you lost me at some point. Are we still talking about adding `:members:` to `c:autodoc` or something else?

BrunoMSantos:
So, let me use an example perhaps. And bear in mind this wouldn't have to happen in this series anyway!

```c
/** Foo. */

/** Fooer.*/
void fooer(void);

/** Bar. */

/** Barer.*/
void barer(void);
```

Then we'd have:

```rst
.. s/block/whatever/
.. autoblock:: Foo
```

Expanding to:

```rst
Foo.
```

And

```rst
.. autoblock:: Foo
   :members:
```

Expanding to:

```rst
Foo.

.. c:function:: void fooer(void)

   Fooer.
```

And so on. Hope that makes it clear. Basically, create a sort of `automodule` in the way I was hoping for, but one that works with a block / section that is bounded by two top level comments.

jnikula:
> So, let me use an example perhaps. And bear in mind this wouldn't have to happen in this series anyway!

Thanks, now I get it. So it would be a way to group stuff, somewhere between `c:autodoc`/`c:automodule` and individual symbols.

And I believe the connection to this series is that `c:autoblock` without `:members:` would yield the same result as `c:autotext`?

If I got that right, the question becomes, what directive naming would make sense for _this_ series such that adding the `:members:` functionality _later_ still makes sense.

Or do we care that if this was two separate directives?

BrunoMSantos:
Yup, I think you got my meaning ;)

> If I got that right, the question becomes, what directive naming would make sense for this series such that adding the :members: functionality later still makes sense.

That would be my only concern assuming you see this as a valid end goal. And yes, I'd much prefer a single directive.

The only API we might want / need to break in the future would be how to identify the relevant top level comment, but I agree with your tradeoff so far.

jnikula:
> That would be my only concern assuming you see this as a valid end goal. And yes, I'd much prefer a single directive.

I think it's valid enough to take it into account now if we can, even if I'm not ready to implement it now.

That said, I don't think it'll much alleviate the folks who want Doxygen `@group`.

> The only API we might want / need to break in the future would be how to identify the relevant top level comment, but I agree with your tradeoff so far.

Thanks.

Now it's just a small matter of figuring out the name for the directive. The "NP" in "NP complete" stands for "Naming Problem", right?

jnikula:
I think I might like `c:autosection` because it conveniently matches both ideas. A section of documentation, extracted from a documentation comment, or a section of documented symbols in source.

BrunoMSantos:
Yeah, autosection sounds good to me too. +1

jnikula:
Autosection in #164

Abdalla123i:
Can we make one for me Android!

PR #163, opened 2023-07-02
Process docstring naming

jnikula:
Move towards "process docstring" naming instead of "transform", folloing the hawkmoth-process-docstring. No functional changes.


BrunoMSantos:
Looks good too. Sorry for taking a bit longer to get to this one.

PR #164, opened 2023-07-02
Autosection

jnikula:
#161 with `c:autosection` and `cpp:autosection` naming, and some tweaks on the documentation/examples, but no real functional changes.


BrunoMSantos:
Looks good to me ;)

Created #166 so we remember the members stuff.

PR #165, opened 2023-07-03
Improve warnings

jnikula:
Our error handling does not seem to be very nice if the file or symbol is not found. Improve them.

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -95,14 +95,26 @@ class _AutoBaseDirective(SphinxDirective):
 
         process_docstring = lambda lines: self.__process_docstring(lines)
 
+        match = False
         for docstrings in root.walk(recurse=False, filter_types=self._docstring_types,
                                     filter_names=self._get_names()):
+            match = True
             for docstr in docstrings.walk(filter_names=self._get_members()):
                 lineoffset = docstr.get_line() - 1
                 lines = docstr.get_docstring(process_docstring=process_docstring)
                 for line in lines:
                     viewlist.append(line, filename, lineoffset)
                     lineoffset += 1
+
+        if not match:
+            if self._get_names():
+                args = ' '.join(self.arguments)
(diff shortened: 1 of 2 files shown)


src/hawkmoth/__init__.py
```diff
@@ -167,7 +167,14 @@ def _get_filenames(self):
                                 location=(self.env.docname, self.lineno))
             return []
 
-        return [os.path.abspath(os.path.join(self.env.config.hawkmoth_root, filename))]
+        filename = os.path.abspath(os.path.join(self.env.config.hawkmoth_root, filename))
+
+        # Warn if the file is not found, but pass it to clang anyway
```
BrunoMSantos, line 184: Hmm, could we let it fail during the parse instead and, in the `except`, figure out the specific cause and adjust the error message?

src/hawkmoth/__init__.py
```diff
@@ -93,15 +93,27 @@ def __parse(self, filename):
...
+        if not match:
+            if self._get_names():
+                args = ' '.join(self.arguments)
+                self.logger.warning(f'"{self.name}:: {args}" does not match documented symbols.',
+                                    location=(self.env.docname, self.lineno))
+            else:
+                # autodoc
+                self.logger.warning('No documented symbols were found.',
+                                    location=(self.env.docname, self.lineno))
+
```
BrunoMSantos, line 116: Nice one!

src/hawkmoth/__init__.py
jnikula: Can do.

jnikula: Force pushed an update

src/hawkmoth/parser.py
```diff
@@ -31,6 +31,7 @@
 """
```
BrunoMSantos, line 31: Typo in commit message: 'fro' -> 'from'.

Thanks for the fix! Looks good to me.

jnikula: Thanks, fixed!

PR #167, opened 2023-07-05
docstring: fix handling of empty documentation comments

jnikula:
Sometimes it's handy to add an empty documentation comment to generate documentation for a symbol without having to add some placeholder text. Or to include documentation for members/enumerations without documenting the struct/enum.

Currently, empty documentation comments lead to:

  File "docstring.py", line 143, in _remove_comment_markers
    while (not lines[0] or lines[0].isspace()):
               ~~~~~^^^
IndexError: list index out of range

Fix it and add some tests and documentation for it.

### DIFF
--- a/test/c/comment-strip.c
+++ b/test/c/comment-strip.c
@@ -107,3 +107,31 @@ int two_liner(int foo, int bar);
  	 */
 int two_liner_whitespace(int foo, int bar);
 
+/**/
+struct empty_comment_0 {
+	/** The embedding struct is not documented. */
+	int description;
+};
+
+/***/
+
+/***/
+struct empty_comment_1 {
+	/** The canonical empty documentation comment. */
+	int description;
+};
+
+/** Top level comment. */
+
+/** */
+struct empty_comment_2 {
+	/** Another empty documentation comment. */
+	int description;
+};
+
+/**
+ */
+struct empty_comment_3 {
+	/** Another empty documentation comment. */
+	int description;
+};
--- a/src/hawkmoth/docstring.py
+++ b/src/hawkmoth/docstring.py
@@ -140,10 +140,10 @@ class Docstring():
         prefix_len = _get_prefix_len(lines[1:-1])
         lines[1:-1] = [line[prefix_len:] for line in lines[1:-1]]
 
-        while not lines[0] or lines[0].isspace():
+        while lines and (not lines[0] or lines[0].isspace()):
             del lines[0]
 
-        while not lines[-1] or lines[-1].isspace():
+        while lines and (not lines[-1] or lines[-1].isspace()):
             del lines[-1]
 
     @staticmethod
--- a/test/c/autosection-comment-strip.rst
+++ b/test/c/autosection-comment-strip.rst
@@ -1,0 +1,3 @@
+
+Top level comment.
+
--- a/test/c/autosection-comment-strip.yaml
+++ b/test/c/autosection-comment-strip.yaml
@@ -1,0 +1,7 @@
+domain: c
+directive: autosection
+directive-arguments:
+  - Top level comment
+directive-options:
+  file: comment-strip.c
+expected: autosection-comment-strip.rst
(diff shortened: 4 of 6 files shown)


doc/syntax.rst
```diff
@@ -46,6 +46,17 @@ includes macros, functions, struct and union members, enumerations, etc.
...
+----------------------------
+
+Use the empty documentation comment ``/***/`` to only generate the C or C++
+Domain directives without documentation. This can be useful for documenting
+struct members or enumeration constants without actually documenting the
+enclosing struct or enum.
+
+It's also a way to let Sphinx know about types to highlight them in
+e.g. parameter documentation.
+
```
BrunoMSantos, line 59: I wonder whether this is necessary at all, but I don't object to it.

test/c/comment-strip.c
```diff
@@ -107,3 +107,27 @@ int two_liner(int foo, int bar);
...
+	/** Another empty documentation comment. */
+	int description;
+};
+
+/**
+ */
+struct empty_comment_3 {
+	/** Another empty documentation comment. */
+	int description;
+};
```
BrunoMSantos, line 133: What about top level comments? Any chance it might be an issue for `autosection`? (I plead ignorance by the way.)

BrunoMSantos:
Interesting edge case. I think it's good over the covered cases and half expecting my worry about top level comments is unfounded. In which case, looks good to me.

Out of curiosity, is this just a pile of things you had laying around or is it being currently driven by actual use cases / requests?

jnikula:
> Interesting edge case. I think it's good over the covered cases and half expecting my worry about top level comments is unfounded. In which case, looks good to me.
> 
> Out of curiosity, is this just a pile of things you had laying around or is it being currently driven by actual use cases / requests?

I was hacking on Mesa documentation a bit, see https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/23145#note_1985368

There were cases where Doxygen+Breathe happily generated documentation if struct members were documented but the struct itself was not. I'm not sure I like to promote that model in Hawkmoth, but I thought adding an empty documentation comment to the enclosing struct would be a nice workaround instead of having to add placeholder text (especially because I was unable to come up with *actual* documentation in the conversion). And it promptly crashed. :facepalm:

The warning improvements came from there as well. There were directives with typoed/wrong names or filenames, and there was no warnings, just silently no documentation.

doc/syntax.rst
jnikula: I can leave that part out. I was contemplating leaving out the entire documentation on empty comments, but thought there's no harm.

test/c/comment-strip.c
jnikula: Did not try yet. But seems like it could be a **feature**! You could use it to end a section (the future version with `:members:` support).

doc/syntax.rst
BrunoMSantos: I meant the whole section too by the way. I see no point in documenting this at all, I think anyone would expect this behaviour intuitively.

test/c/comment-strip.c
BrunoMSantos: Ahahah, maybe...

I'm expecting it to work fine right now, but I'd welcome a unit test for selecting sections before and after one of these at least. In case we inadvertently break it later on.

doc/syntax.rst
jnikula: Maybe just one sentence in the previous section?

"Empty documentation comments `/***/` create the C or C++ Domain directives without documentation."

Or nothing?

BrunoMSantos: Up to you... I don't feel it's needed but I'd prefer a single sentence to a whole section.

test/c/comment-strip.c
jnikula: Dropped it completely. Can be added later if needed.

doc/syntax.rst
jnikula: Dropped it completely. Can be added later if needed.

test/c/comment-strip.c
BrunoMSantos: Yeah, feels complete now. Thanks.

Issue #157, opened 2023-05-24
Make :file: directive option actually optional


jnikula:
Provide a way to reduce or drop the need to specify the file with `:file:` for every directive, every time.

From https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/23145#note_1920081 and https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/23145#note_1923559

"Having to state the file every time is a bit annoying but ultimately harmless."

"Would it work to have a declaration thing where you list a bunch of files and then anything from any of those files can be used without an explicit file declaration?  Or does it really need to be one file at a time?"

#156 picks up the low-hanging fruit, but it might be better to aim higher from the start.

jnikula:
Ping @BrunoMSantos

jnikula:
If source files are specified globally in `conf.py`, they all need to be parsed up front. All rst documents with autodoc directives would need to depend on *all* of the source files, because you don't really know where the referenced symbols come from. I don't think it's enough to note the dependency based on where the symbol was found on one run.

BrunoMSantos:
Nice to see mesa guys reporting their experience with this! I haven't had time to read your RFC yet, so caveat emptor.

Using a directive to select a file makes the documentation directives stateful and dependent on where in the document they appear. So I don't really like that, and unless we can only select a single file, it only reduces the scope of the issues, it doesn't avoid them :/

Performance will indeed take a hit, especially if we want to be accurate (e.g. warning of aliased symbols implies searching all files for each symbol), but I'm not too worried about that honestly, especially if we build a permanent cache on disk. But you haven't touched on the worst bit yet I think.

Currently allow different clang options for different directives, which implies parsing all files for each set of clang options we find (like in #75 actually, though I don't think this issue was discussed then). But that may break things if not all files can compile with a certain set of options, and it's unacceptably expensive to parse a whole project as many times as the user specifies the `:clang:` option.

From the top of my head, we would need to define clang arguments per file / directory (presumably matching those used to build the actual project) and parse files with them. When directives specify additional arguments though, I frankly see no other way around it as we don't know which file we need to parse. Any simplification would come with an accuracy cost I think, so `:clang:` should probably be dropped or issue a loud suggestion that the user specifies `:file:` with it.

I'll think about this more carefully in the weekend.

BrunoMSantos:
> From the top of my head, we would need to define clang arguments per file / directory (presumably matching those used to build the actual project)

[...]

PR #168, opened 2023-07-08
Optional file option

jnikula:
First stab at better separation of parsing and incorporating documentation, and making the file option actually optional.

With this, you can do

```
.. c:autofunction:: foo
   :file: baz.c

.. c:autofunction:: bar
```

And the later directives will match symbols against all previously parsed files in the rst document.

Additionally, there's:

```
.. c:autodoc:: bar.c
   :parse-only:

.. c:autofunction:: bar
```

And it works the same way, but has a more explicit feeling to it.

This is the first step; later I imagine also being able to parse files *before* any rst documents are read, and the symbols are searched against all of them.

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -72,7 +72,7 @@ class _AutoBaseDirective(SphinxDirective):
         key = (filename, tuple(clang_args))
 
         if key in parsed_files:
-            return parsed_files[key]
+            return
 
         # Tell Sphinx about the dependency
         self.env.note_dependency(filename)
@@ -83,8 +83,6 @@ class _AutoBaseDirective(SphinxDirective):
         self.__display_parser_diagnostics(errors)
 
         parsed_files[key] = docstrings
-
-        return docstrings
 
     def __parsed_files(self, filter_filenames=None, filter_clang_args=None):
         parsed_files = self.env.temp_data.get('hawkmoth_parsed_files', {})
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -151,6 +151,9 @@ class _AutoBaseDirective(SphinxDirective):
             for filename in self._get_filenames():
                 self.__parse(filename)
 
+        if 'parse-only' in self.options:
+            return []
+
         result = ViewList()
 
         self.__get_docstrings(result)
@@ -168,6 +171,11 @@ class _AutoDocDirective(_AutoBaseDirective):
     # Allow passing a variable number of file patterns as arguments
     required_arguments = 1
     optional_arguments = 100   # arbitrary limit
+
+    option_spec = _AutoBaseDirective.option_spec.copy()
+    option_spec.update({
+        'parse-only': directives.flag,
+    })
 
     def _get_filenames(self):
         ret = []
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -91,9 +91,7 @@ class _AutoBaseDirective(SphinxDirective):
 
         self.env.app.emit('hawkmoth-process-docstring', lines, transform, self.options)
 
-    def __get_docstrings(self, viewlist, filename):
-        root = self.__parse(filename)
-
+    def __get_docstrings_for_root(self, viewlist, root):
         process_docstring = lambda lines: self.__process_docstring(lines)
 
         match = False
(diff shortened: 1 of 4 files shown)


BrunoMSantos:
Hi Jani, so I haven't read the patches and I'm afraid I can't make any promises until mid August as I'm helping with an event which will eat most of my free time. But I can leave a few comments on your summary.

> And the later directives will match symbols **against all previously parsed files in the rst document**.

Initially I thought this was just a fleshed out version of the stateful 'select the active file' directives, but I see this version is actually much closer to what I'd prefer myself, so I'm glad you're taking that route!

Still, I would like to have a second look at the difficulties in building a single cache by white/blacklisting files and directories in `conf.py`. Is that too far off, you think? Or is that where our views differ?

In order to keep all relevant discussions linked, issue #157 and the abandoned PRs #75 and #156 all have interesting titbits in my view. Some of them not fully discussed and still very relevant I think.

jnikula:
> Hi Jani, so I haven't read the patches and I'm afraid I can't make any promises until mid August as I'm helping with an event which will eat most of my free time. But I can leave a few comments on your summary.

Thanks for the heads up, and the high level comments!

I value your input, and it's always contributed to making the end result better, but I'm not sure I can wait that long with this. I'll try not to make a mess of it. :grimacing: 

> > And the later directives will match symbols **against all previously parsed files in the rst document**.
> 
> Initially I thought this was just a fleshed out version of the stateful 'select the active file' directives, but I see this version is actually much closer to what I'd prefer myself, so I'm glad you're taking that route!

Agreed!

> Still, I would like to have a second look at the difficulties in building a single cache by white/blacklisting files and directories in `conf.py`. Is that too far off, you think? Or is that where our views differ?

I think defining the files one way or the other in `conf.py` should be the end goal, but I think trying to get there right away is too big of a leap at once. For one thing, figuring out when a re-parse is required and storing the parse results in Sphinx cache is non-trivial, at least to me. Matching names against files parsed in the current document is the intermediate step, and does not block the end goal.

The question is *how* the current document scope should work. With the proposal, all previous directives with `:file:` add to the stash of files where symbols will be looked up in for the subsequent directives, if they lack the `:file:` option.

I know this has an implicit feeling to it that you would dislike. But requiring an explicit "hey, parse this file" directive and *preventing* the implicit from matching is actually harder. (Well, it needs a flag saying where the parse results originated from, and only using them if they were explicit.)

The series does add the explicit "hey, parse this file" mechanism via the `:parse-only:` option to `c:autodoc`. But I'm not convinced overloading `c:autodoc` is a good idea. But what would a separate, dedicated directive be?

Maybe we could skip the explicit or dedicated directive, and live with the implicit way with the goal of defining the files in `conf.py`?

> In order to keep all relevant discussions linked, issue #157 and the abandoned PRs #75 and #156 all have interesting titbits in my view. Some of them not fully discussed and still very relevant I think.

I'll re-read them, thanks.

jnikula:
PS. Even if we decide to overload `c:autodoc`, I'm not fond of `:parse-only:`. It's really just a placeholder name I threw in, and as a user visible name conveys too much about the inner workings that is just unnecessary.

BrunoMSantos:
> I value your input, and it's always contributed to making the end result better, but I'm not sure I can wait that long with this. I'll try not to make a mess of it.

Yeah, I get that. I'll try to keep an eye on this nonetheless, so as a rule of thumb, let PRs live for two or three days if you don't mind, I may sneak a peek or two. Besides this is not the 1st time we're drafting or discussing this subject.

> [...] Maybe we could skip the explicit or dedicated directive, and live with the implicit way with the goal of defining the files in conf.py?

Funny, that was my very 1st thought actually. That is the one bit that probably wouldn't carry over to the other solution. But it is ironic indeed that we would be removing the (arguably) better of the two _current_ options. Honestly I could go either way, I'll trust your judgment.

At a 1st glance, a dedicated directive seems worse than the `:parse-only:` to me at least. As for the option name, maybe `:index:` or `:cache:`? Dunno. I don't dislike yours that much to be honest.

> > [...] Some of them not fully discussed and still very relevant I think.
> 
> I'll re-read them, thanks.

I should have added that the relevance may not be not strictly in the context of this implementation. But still, good to keep the whole context in mind ;)

jnikula:
I pushed a commit on top documenting the `:file:` option.

BrunoMSantos:
I've read most of it, maybe not as carefully as I should, but I think it's only missing some tests perhaps. I'll leave the `:parse-only:` for you to decide.

src/hawkmoth/__init__.py
```diff
@@ -160,6 +160,7 @@ class _AutoDocDirective(_AutoBaseDirective):
     optional_arguments = 100   # arbitrary limit
```
BrunoMSantos, line 174: You can always do `list(_get_filenames())` where needed and allow for a marginally better performance and semantics in the cases where it's not needed.

Though this is a drive by comment and it's very possible this is still the way to go. It was the commit message that caught my eye.

jnikula:
v2 #171 merged.

PR #169, opened 2023-07-13
Optional file option prep work

jnikula:
This is all the prep work from #168 without the actual functional change of making the `:file:` option optional. I also dropped the generator removal change, and used `list()` around the caller instead.

I figured this one doesn't really need specific tests, while I figure out how to test the rest of the changes.


BrunoMSantos:
Looks good! Good idea to split it too.

jnikula:
> Looks good! Good idea to split it too.

Thanks for taking the time to look at this!

Issue #175, opened 2023-08-06
Add enumerator values in documentation


jnikula:
The Sphinx C and C++ domain directive for enumerators allows optionally specifying the enumerator value:

```
.. c:enumerator:: name = value
```

It would be straightforward to unconditionally add it (the Cursor for `CursorKind.ENUM_CONSTANT_DECL` has `enum_value` property), but the question is whether that is the right thing to do.

The alternatives:

- Always display value
- Display value if the source code explicitly sets it. This is what Doxygen appears to do. I think this requires looking at the tokens. I've implemented this in #176 
- Make it configurable, could be done globally and at `c:autoenum::`
- Never display value (status quo)

I'm not really fond of adding a lot of configuration, and would prefer sane defaults instead.

@BrunoMSantos thoughts?

BrunoMSantos:
I replied in #176, but to be explicit, I like the Doxygen behaviour here. I don't see much value in making it configurable personally and we can always add that later if anyone wants it I guess, keeping the Doxygen behaviour as default.

Issue #174, opened 2023-07-29
Use symbolic names for array sizes in generated documentation


jnikula:
Given:

```
/** size */
#define SIZE 5

/** array */
int array[SIZE];

```
it would be nice if the generated documentation actually had `int array[SIZE];` instead of the current `int array[5];`

BrunoMSantos:
So, this immediately made me think of #92. Seems like a similar problem asking for a general solution. Same for the enumerator initializers which I just mentioned in #176 and just about any other pre-processor symbol we might not want to expand. Thoughts?

jnikula:
> So, this immediately made me think of #92. Seems like a similar problem asking for a general solution. Same for the enumerator initializers which I just mentioned in #176 and just about any other pre-processor symbol we might not want to expand. Thoughts?

Do you basically mean #135 enhanced with passing the Cursor to the event handler?

For the most part I really just want sane defaults. But beyond that, perhaps the right thing is to give the user the tools to go wild, instead of developing complex configuration options. I don't know. It also seems scary to give the user access to the Cursor, as it creates a dependency on Clang details which we can't hide from the user.

I just merged #176 as the sane default thing to do. But it's prints the numeric value, not the symbolic thing in source.

BrunoMSantos:
> > So, this immediately made me think of #92. Seems like a similar problem asking for a general solution. Same for the enumerator initializers which I just mentioned in #176 and just about any other pre-processor symbol we might not want to expand. Thoughts?
> 
> Do you basically mean #135 enhanced with passing the Cursor to the event handler?

Yeah, I think we've discussed it in the past in fact. Though I'm not sure it should be the 'clang' cursor instead of a documentation centric abstraction of the cursor that allows for easier manipulation by the user. E.g. with methods to extract qualifiers or whatever hiding away all the AST's oddities. But those are details I think.

> For the most part I really just want sane defaults. But beyond that, perhaps the right thing is to give the user the tools to go wild, instead of developing complex configuration options.

Yup, I think the extension is a nice compromise. Whether replacing `_Bool` is the 'right' default or not I'm not so sure. I can think of other cases where it would make about as much sense (e.g. `#define memcpy __builtin_memcpy`) and yet I would prefer it to be expanded by default I think. Consistency and so on...

> I don't know. It also seems scary to give the user access to the Cursor, as it creates a dependency on Clang details which we can't hide from the user.

[...]

PR #176, opened 2023-08-06
Enumerator values

jnikula:
Implement #175


BrunoMSantos:
I like the default, I only think this may benefit from not expanding pre-processor constants, at least in some cases, kind of like in #174 for array sizes. Can be seen as an improvement though, so I wouldn't block it as it currently stands. Good to go!

jnikula:
> I like the default, I only think this may benefit from not expanding pre-processor constants, at least in some cases, kind of like in #174 for array sizes. Can be seen as an improvement though, so I wouldn't block it as it currently stands. Good to go!

Agreed! Thanks!

PR #178, opened 2023-08-13
Remove docker build targets, dockerfiles and stop doing docker releases

jnikula:
Nobody seems to need them.


jnikula:
The dockerfile for running tests still remains.

jnikula:
@BrunoMSantos I presume you're not using them either, correct?

BrunoMSantos:
> @BrunoMSantos I presume you're not using them either, correct?

No I don't and I agree, there's no reason to have this :+1:

PR #184, opened 2023-10-05
parser: workaround cursor.get_tokens() issue with macro expansions

jnikula:
There's a bug in clang that causes `cursor.get_tokens()` to sometimes fail when there are macro expansions in the cursor's extent. Just having a `bool x` variable or struct member fails to tokenize, because bool is a macro rather than a typedef in stdbool.h.

I haven't been able to come up with a reproducer that would hit this with our current usage of `cursor.get_tokens()`. However, it's obvious when trying to tokenize said `bool x`. Rather than wait for us hitting subtle bugs, defensively work around the issue.

The issue seems to be related to something going awry in the cursor's extent. Simply recreating the extent works. The resulting new extent has the same `__repr__`, but they differ under the hood. The new one works, the old one doesn't. Try the regular cursor.get_tokens() first, and if that produces no tokens, fall back to recreating the extent, and getting the tokens from the translation unit.

This is likely caused by clang issue [1], or it's closely related. See also [2].

[1] https://github.com/llvm/llvm-project/issues/43451
[2] https://stackoverflow.com/questions/16786767/obtain-original-unexpanded-macro-text-using-libclang

### DIFF
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -38,6 +38,7 @@ from clang.cindex import TokenKind, CursorKind, TypeKind
 from clang.cindex import StorageClass, AccessSpecifier, ExceptionSpecificationKind
 from clang.cindex import Index, TranslationUnit, TranslationUnitLoadError
 from clang.cindex import Diagnostic
+from clang.cindex import SourceLocation, SourceRange
 
 from hawkmoth import docstring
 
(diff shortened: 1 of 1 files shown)


jnikula:
The fallback is actually not enough. In some cases you can get excessive bogus tokens from `cursor.get_tokens()`. I'll need to update.

src/hawkmoth/parser.py
```diff
@@ -185,6 +186,27 @@ def _comment_extract(tu):
...
     return top_level_comments, comments
 
+# Workaround for clang cursor.get_tokens() being unreliable for cursors whose
+# extent contains macro expansions. The result may be empty or contain bogus
+# tokens, depending on the case.
+#
+# The problem seems to be related to cursor.extent. Recreating the extent and
+# getting the tokens from the translation unit works fine. The __repr__ for both
+# the recreated and original extents is the same, but comparison indicates they
+# do differ under the hood.
```
BrunoMSantos, line 196: We've used this pattern before, but increasingly I think we should place these kind of comments as docstrings of the symbol itself, because why not? Anyway, not really an issue so it's up to you.

BrunoMSantos (approved):
Odd one. Good you found it!

src/hawkmoth/parser.py
jnikula: I think I'll skip this for for now, but overall I agree. And really, the only way for the docstrings to be useful is to actually incorporate them into developer documentation. Not sure if it needs to be part of the regular `make html` build, or maybe hidden behind some knob.

BrunoMSantos: Well, not necessarily... If you were doing something with this on a console as a dev, you could `help('symbol_name')` and such. But it's totally fine.

jnikula: That's just a testament to the fact that I never use `help` like that... and I probably should!

PR #185, opened 2023-10-05
Bool conversion

jnikula:
After finding the reason for my tokenization difficulties (#184) I decided to have another look at the `_Bool` to `bool` conversion.

Long story short, I've had a change of heart.

Let's keep it simple, and just unconditionally convert `_Bool` to `bool`. That's what folks want, that's the C++ and C23 standard, no need to overcomplicate it, or hold it hostage to some new Hawkmoth Sphinx event extension with cursors passed and bells and whistles. Just do it and move on.

### DIFF
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -401,6 +401,9 @@ def _get_scopedenum_type(cursor):
             return f': {cursor.enum_type.spelling}'
     return None
 
+def _normalize_type(type_string):
+    return 'bool' if type_string == '_Bool' else type_string
+
 def _var_type_fixup(cursor, domain):
     """Fix non trivial variable and argument types.
 
@@ -455,7 +458,7 @@ def _var_type_fixup(cursor, domain):
         if len(args) == 0:
             args.append('void')
 
-        ret_type = cursor_type.get_result().spelling
+        ret_type = _normalize_type(cursor_type.get_result().spelling)
 
         name = f'''{pad(ret_type)}({pad(stars_and_quals)}{cursor.spelling}{dims})({', '.join(args)})'''  # noqa: E501
     else:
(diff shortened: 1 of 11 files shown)


src/hawkmoth/parser.py
```diff
@@ -401,6 +401,9 @@ def _get_scopedenum_type(cursor):
             return f': {cursor.enum_type.spelling}'
     return None
 
+def _convert_bool(type_string):
```
BrunoMSantos, line 404: I suggest a more generic `_normalize_type` or something. It reads better in its call site, the fact that it only touches the bool case is irrelevant there.

BrunoMSantos (approved):
Not sure I like that if the code uses `_Bool` explicitly it will be converted as well even though I can hardly conceive of anyone taking notice of that (I wouldn't).
I think it's ok.

jnikula:
> Not sure I like that if the code uses `_Bool` explicitly it will be converted as well even though I can hardly conceive of anyone taking notice of that (I wouldn't). I think it's ok.

So I took this into account first, after I figured out the whole clang tokenization mess. Then I added the tests both for converting `bool` and preserving `_Bool`, and realized `stdbool.h` really has `#define _Bool bool` i.e. the reverse for C++. The preserve test would have had to be different for C++, or we would have had to do the conversion also in the other direction. 

And it was at this point that I decided this is not worth all the complications... just keep it simple.

src/hawkmoth/parser.py
jnikula: Agreed.

jnikula: Fixed.

BrunoMSantos:
> [...] and realized `stdbool.h` really has `#define _Bool bool` i.e. the reverse for C++.

I was just checking it and it's indeed way messier than I thought... Mild objection fully retracted xD

jnikula:
> > [...] and realized `stdbool.h` really has `#define _Bool bool` i.e. the reverse for C++.
> 
> I was just checking it and it's indeed way messier than I thought... Mild objection fully retracted xD

As always, many thanks for the reviews!

PR #186, opened 2023-10-06
Array symbolic dimensions

jnikula:
Things like

```
#define SIZE 5
int foo[SIZE];
```

get documented with the actual value 5 rather than the symbolic `SIZE`. Fix it.

(This includes #184 to ensure it works at all.)

### DIFF
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -426,9 +426,26 @@ def _get_scopedenum_type(cursor):
 def _normalize_type(type_string):
     return 'bool' if type_string == '_Bool' else type_string
 
+def _symbolic_dims(cursor):
+    dim = None
+    for spelling in [t.spelling for t in _cursor_get_tokens(cursor)]:
+        if spelling == '[':
+            # dim should be None here
+            dim = []
+        elif spelling == ']':
+            # dim should not be None here
+            yield ' '.join(dim)
+            dim = None
+        elif dim is not None:
+            dim.append(spelling)
+
 def _dims_fixup(cursor, dims):
     if not dims:
         return ''
+
+    symbolic_dims = list(_symbolic_dims(cursor))
+    if len(symbolic_dims) == len(dims):
+        dims = symbolic_dims
 
     return ''.join([f'[{d}]' for d in dims])
 
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -426,6 +426,12 @@ def _get_scopedenum_type(cursor):
 def _normalize_type(type_string):
     return 'bool' if type_string == '_Bool' else type_string
 
+def _dims_fixup(cursor, dims):
+    if not dims:
+        return ''
+
+    return ''.join([f'[{d}]' for d in dims])
+
 def _var_type_fixup(cursor, domain):
     """Fix non trivial variable and argument types.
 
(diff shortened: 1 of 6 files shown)


src/hawkmoth/parser.py
```diff
@@ -423,10 +423,27 @@ def _get_scopedenum_type(cursor):
...
+    dim = None
+    for spelling in [t.spelling for t in _cursor_get_tokens(cursor)]:
+        if spelling == '[':
+            # dim should be None here
+            dim = []
+        elif spelling == ']':
+            # dim should not be None here
+            yield ' '.join(dim)
+            dim = None
+        elif dim is not None:
```
BrunoMSantos, line 436: This is really just an `else:` right? I think it's better to do it that way. If due to some bug it tries to append something to `None` it will fail loudly instead of continuing silently.

Moreover, at 1st glance this made it look like the `dims` has a secondary function here as a way of signalling where we are in the token processing when in fact it isn't: we're either at the beginning `[`, the end `]`, or it's something to append. I think? I'm tired, so... :sweat_smile:

BrunoMSantos (approved):
There's one comment, but I don't think the behaviour is wrong if you don't address it. Nice one!

src/hawkmoth/parser.py
jnikula: `dims` actually does have that secondary function! The tokens also contain stuff *before* the first `[`, and we want to ignore them. A quick debug print reveals it:

```
diff --git a/src/hawkmoth/parser.py b/src/hawkmoth/parser.py
index 49bcfe96db2a..a9b08cb8ad4c 100644
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -427,6 +427,8 @@ def _normalize_type(type_string):
     return 'bool' if type_string == '_Bool' else type_string
 
 def _symbolic_dims(cursor):
+    print([t.spelling for t in _cursor_get_tokens(cursor)])
+
     dim = None
     for spelling in [t.spelling for t in _cursor_get_tokens(cursor)]:
         if spelling == '[':

```

Running `hawkmoth test/c/array-symbolic-dimensions.c` with that on top produces e.g.:

```
['static', 'int', 'foo', '[', 'MACRO_SIZE', ']']
['static', 'int', 'bar', '[', 'MACRO_SIZE', '*', 'ENUM_SIZE', ']', '[', '3', ']']
```

jnikula:
The force push was just a rebase because there were conflicts.

src/hawkmoth/parser.py
BrunoMSantos: Ah, right! I guess I should have tested it.

Not sure I'm a fan, it may be a bit too clever. But tests are failing on my side (probably some update, I hadn't picked this up in a while), so I won't be volunteering any alternative right now.

It's quite readable anyway, it's :+1: from me.

jnikula: I'm always eager to merge after seeing the thumbs up... but just double checking, by "a bit too clever" you mean the details of the implementation (which can be improved later), not the idea?

BrunoMSantos: Hey, sorry I thought I had replied to this. Yeah, I meant the implementation with regards to the use of the `dims` variable and such. But again, it's totally readable so I think it's fine.

jnikula: Merged now anyway, thanks!

PR #188, opened 2023-10-07
Add links to externally hosted source code for documented symbols

jnikula:
Add `[source]` links for documented symbols, similar to what sphinx.ext.viewcode and sphinx.ext.linkcode do.

https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html
https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html

Running `make html` and looking at examples.html gives the best idea what this does.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -61,6 +61,11 @@ if 'READTHEDOCS' in os.environ:
 
 hawkmoth_root = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../test/examples')
 
+source_uri = 'https://github.com/jnikula/hawkmoth/tree/{version}/test/examples/{{source}}#L{{line}}'
+source_version = f'v{version}' if len(version.split('.')) == 3 else 'master'
+
+hawkmoth_source_uri = source_uri.format(version=source_version)
+
 # -- Options for HTML output -------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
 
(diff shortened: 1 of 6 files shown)


doc/conf.py
```diff
@@ -56,6 +56,11 @@
 
 hawkmoth_root = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../test/examples')
 
+source_uri = 'https://github.com/jnikula/hawkmoth/tree/{version}/test/examples/{{source}}#L{{line}}'
+source_version = version if len(version.split('.')) == 3 else 'master'
```
BrunoMSantos, line 60: Wondering if we should have an intermediate fallback to a specific hash revision instead of using a branch name which is moving target. But I guess it's fine for the current use case.

BrunoMSantos (approved):
LGTM ;)

jnikula:
I noticed some weirdness in the line numbers, also in warnings. Probably related to how docstring adds/removes lines. I'll look into it before merging this one.

doc/conf.py
jnikula: That requires looking at git, and opens up more problems. I'd revisit that later.

I did notice the VERSION file does not have v prefix, but the tags do, so I fixed that.

jnikula:
> I noticed some weirdness in the line numbers, also in warnings. Probably related to how docstring adds/removes lines. I'll look into it before merging this one.

Added a new commit fixing that, and adjusted signode.line a bit. The references seem more sensible now.

BrunoMSantos:
Cool. Looks even better now I guess :)

jnikula:
> Cool. Looks even better now I guess :)

Thanks, merged!

Issue #189, opened 2023-10-08
Test failures on Fedora linux (Clang 16)


bkmgit:
The package builds, but there seem to be problems running the tests. For example:
```
+ cd hawkmoth-0.15.0
+ pushd test
~/build/BUILD/hawkmoth-0.15.0/test ~/build/BUILD/hawkmoth-0.15.0
+ CFLAGS='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Werror=implicit-function-declaration -Werror=implicit-int -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m32 -march=i686 -mtune=generic -msse2 -mfpmath=sse -mstackrealign -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection '
+ LDFLAGS='-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1  '
+ PATH=/builddir/build/BUILDROOT/python-hawkmoth-0.15.0-1.fc40.noarch/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin
+ PYTHONPATH=/builddir/build/BUILDROOT/python-hawkmoth-0.15.0-1.fc40.noarch/usr/lib/python3.12/site-packages:/builddir/build/BUILDROOT/python-hawkmoth-0.15.0-1.fc40.noarch/usr/lib/python3.12/site-packages
+ PYTHONDONTWRITEBYTECODE=1
+ PYTEST_ADDOPTS=' --ignore=/builddir/build/BUILD/hawkmoth-0.15.0/.pyproject-builddir'
+ PYTEST_XDIST_AUTO_NUM_WORKERS=5
+ /usr/bin/pytest -k test_cli .
============================= test session starts ==============================
platform linux -- Python 3.12.0, pytest-7.4.2, pluggy-1.2.0
rootdir: /builddir/build/BUILD/hawkmoth-0.15.0/test
configfile: pytest.ini
[...]

PR #190, opened 2023-10-08
parser: fix parsing of anonymous entities with clang 16+

jnikula:
Clang 16 started returning names for anonymous entities [1], of the form:

(unnamed <entity> at /path/to/file:<line>:<column>)

Additionally check for is_anonymous().

[1] https://github.com/llvm/llvm-project/commit/19e984ef8f49bc3ccced15621989fa9703b2cd5b

---

I'm not sure if this actually fixes anything, because I didn't try to install clang 16 yet. But at least this doesn't break anything on my clang 14.

Maybe a fix to #189.

### DIFF
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -507,8 +507,14 @@ def _type_definition_fixup(cursor):
     type_elem = []
 
     # Short cut for anonymous symbols.
-    if cursor.spelling == '':
+    if cursor.is_anonymous():
         return None
+
+    # libclang 16 and later have cursor.spelling == cursor.type.spelling for
+    # typedefs of anonymous entities, while libclang 15 and earlier have an
+    # empty string. Match the behaviour across libclang versions.
+    if cursor.spelling == '':
+        return cursor.type.spelling
 
     type_elem.extend(_specifiers_fixup(cursor, cursor.type))
 
--- a/test/c/typedef-enum.rst
+++ b/test/c/typedef-enum.rst
@@ -9,7 +9,7 @@
       named enumeration
 
 
-.. c:enum:: @anonymous_90372874a3c8c25dccf983612f39e93f
+.. c:enum:: unnamed_t
 
    unnamed typedeffed enum
 
--- a/test/c/typedef-struct.rst
+++ b/test/c/typedef-struct.rst
@@ -9,7 +9,7 @@
       named member
 
 
-.. c:struct:: @anonymous_7f9a1d628cd33f3227f3fcdc3a405aa6
+.. c:struct:: typedef_struct
 
    unnamed typedeffed struct
 
--- a/test/cpp/typedef-enum.rst
+++ b/test/cpp/typedef-enum.rst
@@ -9,7 +9,7 @@
       named enumeration
 
 
-.. cpp:enum:: @anonymous_90372874a3c8c25dccf983612f39e93f
+.. cpp:enum:: unnamed_t
 
    unnamed typedeffed enum
 
--- a/test/cpp/typedef-struct.rst
+++ b/test/cpp/typedef-struct.rst
@@ -9,7 +9,7 @@
       named member
 
 
-.. cpp:struct:: @anonymous_7f9a1d628cd33f3227f3fcdc3a405aa6
+.. cpp:struct:: typedef_struct
 
    unnamed typedeffed struct
 


jnikula:
This is not enough.

jnikula:
> This is not enough.

Not enough, but it does make progress. It fixes all actually anonymous cases, where clang now returns the "helpful" `(unnamed struct at /src/test/cpp/../c/struct.c:49:1)` kind of names in `cursor.spelling`.

What remains are typedefs of anonymous structs/unions/enums, where the typedef name is now returned as the struct/union/enum name, and `is_anonymous()` returns false.

This might actually be a useful thing... except would have to make it work on pre-16 clang too.

jnikula:
Updated with a proper fix across clang versions.

BrunoMSantos:
So it turns out this was a big part of why my tests were broken too, and it does improve things substantially (I'm on v16). That said, I still have plenty of other test issues. I still haven't gone through it all, but doesn't seem related to anonymous stuff, so I don't think they matter here, but, for the record, most errors seem to come from `bool` showing up as `int`.

Anyway, regarding these changes, the code looks fine. I'm not sure the new behaviour is better than the `anonymous_<hash>` though. Those were unique before and were obviously marked as anonymous type names, we can't say the same thing about the new ones. There's potential for someone referencing one of these unaware that it's not actually a named type they're referencing, maybe even clashing with a type elsewhere that actually bears the same name and that whole mess. Honestly, I'd rather coerce things to the original format across all versions.

BrunoMSantos:
> Honestly, I'd rather coerce things to the original format across all versions.

Or replicate the new behaviour across all versions... That would be even better I guess, though a bit more difficult perhaps.

jnikula:
> Or replicate the new behaviour across all versions... That would be even better I guess, though a bit more difficult perhaps.

That's what the commit is supposed to do, though!

`is_anonymous()` covers all "regular" anonymous entities. Before clang 16, they had empty string for `cursor.spelling`, and since clang 16, there's this "unnamed at" debug style thing.

After that, what remains are the typedeffed anonymous entities. Before clang 16, they also had empty string for `cursor.spelling`, but since clang 16, it's `cursor.type.spelling` - which is what I'm now using for before clang 16 too.

jnikula:
Also, it seems actually harder to unify on the old behaviour. With old clang, you can tell it's a typedeffed anonymous entity from the empty `cursor.spelling`... but I don't know how to detect that case with clang 16!

BrunoMSantos:
Sorry, I meant printing for all versions (e.g.):
```rst
c:enum:: enum (unnamed at /builddir/build/BUILD/hawkmoth-0.15.0/test/c/enum.c:18:1)
```
instead of
```rst
c:enum:: @anonymous_ef849cb791c3c921354e4b05dcefedfa
```

BrunoMSantos:
Ah wait, I think I get it. Never mind me, I'm doing too many things at once... I'll have a closer look later.

src/hawkmoth/parser.py
```diff
@@ -507,9 +507,15 @@ def _type_definition_fixup(cursor):
...
-    if cursor.spelling == '':
+    if cursor.is_anonymous():
         return None
 
+    # libclang 16 and later have cursor.spelling == cursor.type.spelling for
+    # typedefs of anonymous entities, while libclang 15 and earlier have an
+    # empty string. Match the behaviour across libclang versions.
+    if cursor.spelling == '':
+        return cursor.type.spelling
+
```
BrunoMSantos, line 518: So this is indeed tricky. The only way I can see around it is to look into the tokens, which is not great. At the same time I'm not happy with the proposed solution where the typedefs are erroneously documented as structures. It's unexpected and may lead to confusion...

With a quick test (v16), this should work as an alternative with no changes to the unit test results:

```python
    # Short cut for anonymous symbols. (Not sure we need both conditions)
    if cursor.is_anonymous() or cursor.spelling == '':
        return None

    # libclang 16 and later have cursor.spelling == cursor.type.spelling for
    # typedefs of anonymous entities. Confirm that it's an anonymous
    # declaration by ensuring the 2nd token is a '{', e.g. 'struct {...'.
    tokens = cursor.get_tokens()
    next(tokens)
    if next(tokens).spelling == '{':
        return None
```

Not sure how brittle it is (it's quite a shallow approach), but it passes all tests so far.

Regardless of whether the token parsing needs to be more robust, we need to decide whether the output consistency is worth this sort of thing. I'm leaning yes: it solves a real issue and we have several precedents for looking into the tokens when the AST doesn't give us a digested answer by itself. What do you think?

jnikula: > At the same time I'm not happy with the proposed solution where the typedefs are erroneously documented as structures. It's unexpected and may lead to confusion...

As a practical example, when converting Mesa to use Hawkmoth, I've had to give names to a bunch of typedeffed anonymous structs just because of documentation. It's not great, because it's a common pattern, and converting `typedef struct { ... } foo` to `typedef struct foo { ... } foo` feels silly. They are both typedefs, yet they both produce struct documentation, and you still can't reference the typedef in Sphinx in any way. Unless you separate the named struct and the typedef and document them separately.

I think the expectation actually is `typedef struct { ... } foo` documents a struct and lets you reference it as `foo`.

As a data point, Doxygen+Breathe also does this.

BrunoMSantos: > As a practical example, when converting Mesa to use Hawkmoth, I've had to give names to a bunch of typedeffed anonymous structs just because of documentation.

Hmm, that was a conscious decision we took a while back and I still believe it was the right call. There is a difference:

```c
struct some_name {
        ...
};

/** This will be documented with `c:type`, not `c:struct`. */
typedef struct some_name new_name;
```

In the proposed solution, we document something as a struct which isn't. E.g., taking the example from the unit test, one cannot do `struct typedef_struct var;` but can certainly do `struct named var;`, yet the documentation would hint otherwise.

I agree it's a common pattern, but the real solution there would be to make the parser special case these scenarios and use `c:type` instead. I think that's a bit trickier though and I was happy to leave that in the wish list as long as we had expectable and simple behaviour. In fact we did try to do it the smart way before and it was the ensuing discussion over edge cases that led to the decision of sticking to the current behaviour. At the time we actually had opposite roles in the discussion if I remember correctly, but you thoroughly convinced me :sweat_smile: 

As for Doxygen, does it special case it as a type documentation or does it document the typedef as a struct? Just curiosity... I wouldn't really condition any decision here on what Doxygen does.

jnikula: Doxygen+Breathe documents the typedeffed structs as structs, not as types, regardless of whether it's anonymous or not. But hey, so do we! The question here is, what name to give that.

If we were to document such things as types, the user would be forced to separate the struct and typedef both in code and documentation, because you can't have member documentation within a type documentation.

Arguably defaulting to the typedef name is more flexible, because you can use that as the easy default, but you can also separate them and document the struct and type separately, but you're not forced to do that for the common case.

See also https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/24507/diffs?commit_id=fa3684a2caa3a250b1c3fbfaedcc6cdfc7f1ec5e

That's not required for Doxygen+Breathe.

jnikula: > At the time we actually had opposite roles in the discussion if I remember correctly, but you thoroughly convinced me

Heh. In my defense, you've fixed the type fixups and it's all much clearer now. :)

BrunoMSantos: > If we were to document such things as types, the user would be forced to separate the struct and typedef both in code and documentation, because you can't have member documentation within a type documentation

Damn, that's a good point... Have you tried it though? Sphinx seems to eat this up just fine and renders it beautifully as the typedef it is. Did I miss anything?

```rst
.. c:type:: @anonymous_hash some_struct

   Some typedefed struct.

   .. c:member:: int foo

      Foo.

:c:type:`some_struct`

:c:member:`some_struct.foo`
```

> Arguably defaulting to the typedef name is more flexible, because you can use that as the easy default, but you can also separate them and document the struct and type separately, but you're not forced to do that for the common case.

Very true, and I don't want to fight it too much. I said my piece, and this is certainly not about the code, which looks fine. I am not thoroughly convinced this time around though :P

jnikula: We're going to have to let this one simmer for a bit anyway, as I'm taking some time off. :)

BrunoMSantos: > Doxygen+Breathe documents the typedeffed structs as structs, not as types, regardless of whether it's anonymous or not. But hey, so do we! The question here is, what name to give that.

Forgot to reply to this, but for completeness, no we don't! We will do that once we merge this, but currently we document the structure being typedefed as a structure. A structure that happens to be anonymous and so gets an anonymous name.

What you did in Mesa is frankly a poor workaround, even if I get where you're coming from. You're naming a structure because hawkmoth can't contextually document it as a typedef... It works better sure, but you never actually document the typedef which is what you wanted.

BrunoMSantos: And in case it wasn't clear, I'm conceding my position. I just wanted to note my arguments against as best I could.

jnikula:
So I decided to merge this in the end. This is what clang gives us as the name from v16 on, so let's roll with that. If folks want to document typedefs and structs separately, they still can by splitting their definitions.

PR #191, opened 2023-10-08
Docker updates

jnikula:
Refactor the image building a bit to better support multiple Dockerfiles, and add Dockerfiles to run the tests in various versions of Debian and Fedora. They uncover issues with including `stdbool.h` on Fedora, and handling anonymous entities with Clang 16.


jnikula:
@BrunoMSantos I went ahead and merged this to help debug issues. Not part of the more important stuff.

PR #192, opened 2023-10-09
Fix and capture test app build warnings

jnikula:
Fix docutils warnings in `app = Sphinx()`. They're usually not visible, but when hitting other errors, pytest has captured the stderr and emits the warnings. Fix them, capture the stderr, ensure there are no errors with `app = Sphinx()` going forward, and compare the errors against expected errors like the parser tests do.

### DIFF
--- a/test/test_cautodoc.py
+++ b/test/test_cautodoc.py
@@ -8,6 +8,7 @@ import tempfile
 
 import pytest
 from sphinx.application import Sphinx
+from sphinx.util.docutils import docutils_namespace, patch_docutils
 
 from test import testenv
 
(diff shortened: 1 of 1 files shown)


jnikula:
The errors can be seen e.g. with `make docker-test.fedora-37`. It's hitting an issue with `stdbool.h`, but the logs have:

```
----------------------------- Captured stderr call -----------------------------
WARNING: while setting up extension sphinx.addnodes: node class 'toctree' is already registered, its visitors will be overridden
WARNING: while setting up extension sphinx.addnodes: node class 'desc' is already registered, its visitors will be overridden
WARNING: while setting up extension sphinx.addnodes: node class 'desc_signature' is already registered, its visitors will be overridden
WARNING: while setting up extension sphinx.addnodes: node class 'desc_signature_line' is already registered, its visitors will be overridden
WARNING: while setting up extension sphinx.addnodes: node class 'desc_content' is already registered, its visitors will be overridden
WARNING: while setting up extension sphinx.addnodes: node class 'desc_inline' is already registered, its visitors will be overridden
```

And it goes on and on.

BrunoMSantos (approved):
Nice fixes!

PR #194, opened 2023-10-09
test: query system include args from clang

### LINKED ISSUE #193: Python Clang Bindings don't find standard headers on Fedora
Recently, we added `#include <stdbool.h>` to a number of tests. It works fine on various versions of Debian.

However, on Fedora, it's not found. At least not in the recently added Docker images. Run `make docker-test.fedora-37` to reproduce.

Is something missing in the Dockerfiles? Some clang configuration different on Fedora? The header files can be found in `/usr/lib64/clang/15.0.7/include/stdbool.h`.

Running `clang` inside the container to figure out include paths gives:

```
$ cat /dev/null | clang -E -Wp,-v -
clang -cc1 version 15.0.7 based upon LLVM 15.0.7 default target x86_64-redhat-linux-gnu
ignoring nonexistent directory "/usr/bin/../lib/gcc/x86_64-redhat-linux/12/../../../../x86_64-redhat-linux/include"
ignoring nonexistent directory "/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib64/clang/15.0.7/include
 /usr/local/include
 /usr/include
End of search list.
# 1 "<stdin>"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 370 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "<stdin>" 2
```

`hawkmoth test/c/variable.c` inside the container fails with `CRITICAL: test/c/variable.c:1: 'stdbool.h' file not found`.

Explicitly specifying the include path works: `hawkmoth --clang=-I/usr/lib64/clang/15.0.7/include test/c/variable.c`

But Debian does not require that.

jnikula:
Apparently libclang (and consequently Clang Python Bindings) does not universally get its system include paths right by default, or the same as clang. The defaults work on Debian, but not on Fedora, for example, and I've been unable to root cause this.

Anyway, we already have a helper for querying the system include args from clang, so use it.

Fixes #193.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -35,6 +35,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
+      - name: Install dependencies (apt)
+        run: sudo apt-get install -y python3-clang python3-pip
       - name: Examples check
         run: |
           . venv
--- a/test/conf.py
+++ b/test/conf.py
@@ -3,6 +3,8 @@
 # For the full list of built-in configuration values, see the documentation:
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
+from hawkmoth.util import compiler
+
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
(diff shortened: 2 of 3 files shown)


BrunoMSantos:
Yup, this fixes a bunch of issues for me too. Weird, but I see no real downside either. Good to go ;)

Issue #203, opened 2023-10-22
Make it possible to build hawkmoth documentation without the package installed


jnikula:
https://github.com/jnikula/hawkmoth/issues/197#issuecomment-1768442622_ by @kloczek

Would you accept PR with below patch?
```patch
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -5,6 +5,7 @@

 import os
 import sys
+sys.path.insert(0, os.path.abspath('../src'))

 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
```
It allows build documention out of only source tree without have `hawkmoth` installed.
```console
+ /usr/bin/sphinx-build -n -T -b man doc build/sphinx/man
Running Sphinx v7.1.2
making output directory... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://www.sphinx-doc.org/en/master/objects.inv...
building [mo]: targets for 0 po files that are out of date
writing output...
building [man]: all manpages
updating environment: [new config] 9 added, 0 changed, 0 removed
reading sources... [100%] troubleshooting
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
writing... python-hawkmoth.3 { installation extension directives syntax examples extending built-in-extensions troubleshooting } done
build succeeded.

The manual pages are in build/sphinx/man.
```
🤔

_Originally posted by @kloczek in https://github.com/jnikula/hawkmoth/issues/197#issuecomment-1768442622_

jnikula:
https://github.com/jnikula/hawkmoth/issues/197#issuecomment-1773905066

> Wold you accept PR with below patch?
> 
> ```diff
> --- a/doc/conf.py
> +++ b/doc/conf.py
> @@ -5,6 +5,7 @@
> 
>  import os
>  import sys
> +sys.path.insert(0, os.path.abspath('../src'))
> 
>  # -- Project information -----------------------------------------------------
>  # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
> ```
> 
> It allows build documention out of only source tree without have `hawkmoth` installed.

The reason we don't have this in docs or tests is to ensure they all work with the *installed* package. Building the docs is also part of testing in the sense that they dogfood the package in a normal Sphinx setting instead of in a pytest suite.

[...]

PR #204, opened 2023-10-22
allow documentation build using hawkmoth from source tree

jnikula:
Address #203 while also using the installed package for CI and testing. Best of both worlds?


BrunoMSantos (approved):
:+1:

PR #205, opened 2023-10-22
extension: map parser CRITICAL loglevel to Sphinx ERROR

jnikula:
Sphinx does recognize CRITICAL loglevel, but internally it never uses it, and it does not color highlight CRITICAL the way it highlights ERROR.

Map parser CRITICAL to Sphinx ERROR level to get the console highlighting.

See also https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/logging.py#L45


BrunoMSantos (approved):
:+1:

jnikula:
> 👍

Thanks, merged!

PR #207, opened 2023-10-31
Cli updates

jnikula:
Fix a double parse, and add --version option.


BrunoMSantos (approved):
Looks good! Sorry for taking so long to reply... Busy days!
Haven't forgotten about the cursor bit either! Getting there ;)

PR #211, opened 2023-11-09
Add basic support for C++ namespaces

### LINKED ISSUE #209: C++: documented symbols in namespaces are ignored
Any documented symbols in a namespace are currently ignored. This is probably a known bug.

To reproduce, create this `test.hpp`:
```cpp
namespace A {
/**
 * Test
 *
 * @param a a
 * @return a
*/
int test(int a) { return a; }
}
```
Here `hawkmoth --domain cpp test.hpp`  shows no output. However, removing the namespace results in the correct documentation.

stephanlachnit:
Closes #209

I added support for C++ namespace based on #206. It was actually not too hard, it's basically just adding a cursor kind check in `_parse_undocumented_block`. 

The nasty part was working around the `Docstring` class, as it expects a commented cursor. I wondering if it is worth to add a `BasicDocstring` class, which does not require this.

The implementation I chose uses the namespacing directives from sphinx (https://www.sphinx-doc.org/en/master/usage/domains/cpp.html#namespacing), as it is significantly easier than adding the namespace attribute in front of every symbol correctly. This does mean that the namespace is not shown in the documentation at all.

Basically there two things missing:
- [ ] Add an option to add a paragraphs for namespaces (including filters)
- [ ] Ensure that namespace objects are kept when filtering for specific symbol and not the entire file

It should be mentioned that `using namespace XYZ;` is entirely ignored, but I don't think there is a sane way to fix that, and it shouldn't be used in headers anyway.

### DIFF
--- a/src/hawkmoth/doccursor.py
+++ b/src/hawkmoth/doccursor.py
@@ -531,7 +531,12 @@ class DocCursor:
             if child._cc.kind == CursorKind.CXX_BASE_SPECIFIER:
                 def pad(s): return s + ' ' if s else ''
                 access_spec = child._get_access_specifier()
-                inherited.append(f'{pad(access_spec)}{child._cc.type.spelling}')
+                if child._cc.referenced.kind == CursorKind.CLASS_DECL:
+                    # use referenced type if possible for full namespace
+                    spelling = child._cc.referenced.type.spelling
+                else:
+                    spelling = child._cc.type.spelling
+                inherited.append(f'{pad(access_spec)}{spelling}')
 
         return ': ' + ', '.join(inherited) if len(inherited) > 0 else None
 
--- a/src/hawkmoth/doccursor.py
+++ b/src/hawkmoth/doccursor.py
@@ -14,6 +14,23 @@ from clang.cindex import (
 )
 
 
+def _get_semantic_parent_namespace(cursor, namespace):
+    semantic_parent = cursor.semantic_parent
+    if not semantic_parent:
+        return namespace
+
+    if semantic_parent.kind == CursorKind.NAMESPACE:
+        # parent is a namespace => add namespace in front
+        if namespace is not None:
+            namespace = f'{semantic_parent.spelling}::{namespace}'
+        else:
+            namespace = semantic_parent.spelling
+        # check again for nested namespaces
+        return _get_semantic_parent_namespace(semantic_parent, namespace)
+
+    return namespace
+
+
 class DocCursor:
     """Documentation centric wrapper for Clang's own ``Cursor``.
 
@@ -66,7 +83,7 @@ class DocCursor:
 
     @property
     def name(self):
-        return self._cc.spelling if self._cc.spelling else self.decl_name
+        return self.namespace_prefix + self._cc.spelling if self._cc.spelling else self.decl_name
 
     @property
     def decl_name(self):
@@ -81,7 +98,14 @@ class DocCursor:
             return self._type_definition_fixup()
         else:
             # self.name would recurse back here if self._cc.spelling is None
-            return self._cc.spelling
+            return self.namespace_prefix + self._cc.spelling if self._cc.spelling else None
+
+    @property
+    def namespace_prefix(self):
+        if self.domain != 'cpp':
+            return ''
+        namespace = _get_semantic_parent_namespace(self._cc, None)
+        return f'{namespace}::' if namespace else ''
 
     @property
     def type(self):
@@ -270,7 +294,7 @@ class DocCursor:
         template = self._get_template_line()
         template = template + ' ' if template else ''
 
-        return f'{template}{self._cc.spelling}{colon_suffix}'
+        return f'{template}{self.namespace_prefix}{self._cc.spelling}{colon_suffix}'
 
     def _get_macro_args(self):
         """Get macro arguments.
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -321,6 +321,17 @@ def _parse_undocumented_block(errors, cursor, nest):
                 if c.comment:
                     ret.extend(_recursive_parse(errors, c, nest))
 
+    elif cursor.kind == CursorKind.NAMESPACE:
+        # ignore internal STL namespaces
+        if cursor.name in ['std', '__gnu_cxx', '__cxxabiv1', '__gnu_debug']:
+            return ret
+        # iterate over namespace
+        for c in cursor.get_children():
+            if c.comment:
+                ret.extend(_recursive_parse(errors, c, nest))
+            else:
+                ret.extend(_parse_undocumented_block(errors, c, nest))
+
     return ret
 
 def _language_option(filename, domain):


stephanlachnit:
I used the following file for testing:
```cpp
namespace A {
namespace B {

/**
 * Test fct A
*/
int testa(int a) { return a; }

/**
 * Test cls B
 */
class TestB {
public:
    /**
     * Test fct B
    */
   int testb(int b) { return b; }

private:
    /** Not constant */
    double b;
};


/** Some constant */
constexpr double CONSTANT = 5.;

} // namespace B

/**
 * Test cls D
 */
class TestD : public B::TestB {
public:
    /**
     * Test fct D
    */
   int testd() const { return d; }
private:
    static double d;
};

/**
 * Test fct C
*/
template<typename T>
int testc(int c) { return c; }

/**
 * Test enum E
*/
enum class TestE {
    /** enum member A */
    A,
    /** enum member B */
    B,
};

} // namespace A
```

BrunoMSantos:
> Closes #209
> 
> I added support for C++ namespace based on #206. It was actually not too hard, it's basically just adding a cursor kind check in `_parse_undocumented_block`.
> 
> The nasty part was working around the `Docstring` class, as it expects a commented cursor. I wondering if it is worth to add a `BasicDocstring` class, which does not require this.
> 
> The implementation I chose uses the namespacing directives from sphinx (https://www.sphinx-doc.org/en/master/usage/domains/cpp.html#namespacing), as it is significantly easier than adding the namespace attribute in front of every symbol correctly. This does mean that the namespace is not shown in the documentation at all.

Problem with that is that directives are context dependent and so documenting a specific symbol instead of an entire file will not issue the required push/pops.

> Basically there two things missing:
> 
>     * [ ]  Add an option to add a paragraphs for namespaces (including filters)

You mean documenting namespaces? From my reading of Sphinx documentation, that is not the intended purpose of those directives. `cpp:namespace*` may have 'namespace' in the name, but are not actually meant to map to C++ namespaces nor to produce visible namespace documentation blocks in the output. Haven't tried this since a few months ago though, so I may be out of data. They are 'documentation namespaces', though I suppose _most_ people would want to use them in a way that mirror the C++ namespaces.

They're meant to be used, among others, to scope (e.g.) two different code bases or any other random thing like that. So much so that there are also `c:namespace` directives and C has no concept of namespaces.

In fact, I don't think hawkmoth should get anywhere near those directives and leave those for discretionary use by the users. Instead, we should keep track of the namespace stack and append it on every symbol name.

>     * [ ]  Ensure that namespace objects are kept when filtering for specific symbol and not the entire file

Yup, that is the problem indeed.

> It should be mentioned that `using namespace XYZ;` is entirely ignored, but I don't think there is a sane way to fix that, and it shouldn't be used in headers anyway.

I'm hoping these can be solved as well in fact. Clang should know where symbols come from, so it should be ok.

stephanlachnit:
> Problem with that is that directives are context dependent and so documenting a specific symbol instead of an entire file will not issue the required push/pops.

Yes, but that is relatively easy to fix.

> You mean documenting namespaces? From my reading of Sphinx documentation, that is not the intended purpose of those directives. `cpp:namespace*` may have 'namespace' in the name, but are not actually meant to map to C++ namespaces nor to produce visible namespace documentation blocks in the output. Haven't tried this since a few months ago though, so I may be out of date. They are 'documentation namespaces', though I suppose _most_ people would want to use them in a way that mirror the C++ namespaces.

That was my first thought as well, but I don't think this is correct. It mentions

>  Note that the namespace does not *need* to correspond to C++ namespaces, but can [also] end in names of classes
> [...]
> All subsequent objects will be defined as if their name were declared with the scope prepended. The subsequent cross-references will be searched for starting in the current scope.

The wording implies to me that this *can also* be used to to e.g. enter the namespace of a class or an enum besides the usecase for C++ namespace. It's a bit unfortunate that the sphinx documentation isn't quite clear here.

> They're meant to be used, among others, to scope (e.g.) two different code bases or any other random thing like that. So much so that there are also `c:namespace` directives and C has no concept of namespaces.

I think `cpp:namespace` just combines the use of e.g. entering enum "namespaces" that also exist in C, with the idea of C++ namespaces. After all, they are not too different. clangs use of the `CursorKind.NAMESPACE` even matches this definition (it gives you such a cursor e.g. for enums).

> In fact, I don't think hawkmoth should get anywhere near those directives and leave those for discretionary use by the users. Instead, we should keep track of the namespace stack and append it on every symbol name.

That is actually what I tried first. However, this gets really tricky really quickly. For example
```cpp
namespace project{
namespace subcomponent {
class A {};
class B : public A {};
}
}
```
In the declaration A, in which namespaces does the annotation of B live? It's not obvious to me how to figure this out. clang can probably do that, I haven't looked into it.

But even then, it is essential that a global namespace can be ignored in the documentation, as most project have a global project-specific namespace. Thus, you now need a method to annotate class A not just as
```
class project::subcomponent::A : public project::subcomponent::B
```
but also as the much more readable
```
class subcomponent::A : public subcomponent::B
```.

From a user documentation perspective, the namespace do exactly what I want. When I am in a different namespace, I can reference to the object with `subcomponent::A`, while in the same namespace with `A`.

> I'm hoping these can be solved as well in fact. Clang should know where symbols come from, so it should be ok.

That would be really nice, however I think for the time being this is significantly better than nothing. I will test this code later in our project,  and report back how it worked out.

**Comment by BrunoMSantos** [comment]
> [...]
> I think `cpp:namespace` just combines the use of e.g. entering enum "namespaces" that also exist in C, with the idea of C++ namespaces. After all, they are not too different. clangs use of the `CursorKind.NAMESPACE` even matches this definition (it gives you such a cursor e.g. for enums).

Well, names aside, Clang needs to parse namespaces in C++, it's only natural they show up in the syntax tree. But that's a bit of a red herring. I'm not contesting those directives can be used in this way, I'm proposing that it's not the best / easiest way to do it.

Anyway, I should actually get into this again, I'm mostly giving 1st impressions and talking from memory. I'll have a closer look at this for sure ;)
 
> That is actually what I tried first. However, this gets really tricky really quickly. For example
> 
> ```c++
> namespace project{
> namespace subcomponent {
> class A {};
> class B : public A {};
> }
> }
> ```
> 
> In the declaration A, in which namespaces does the annotation of B live? It's not obvious to me how to figure this out. clang can probably do that, I haven't looked into it.

Hmm, not sure what the issue is. It seems very straightforward and actually independent of whether we use the namespace directives or not (think `.. cpp:autoclass:: project::subcomponent::B` or maybe `.. cpp:autoclass:: B` if we can manually issue the namespace directives). Can you elaborate?

> But even then, it is essential that a global namespace can be ignored in the documentation, as most project have a global project-specific namespace. Thus, you now need a method to annotate class A not just as
> [...]

Hmm, that is making assumptions I _strongly_ oppose. I've worked in projects that have multiple 'root' namespaces. In fact, having no 'root' namespace at all is no crime either, even if it may be uncommon, or even a bad idea in some cases.

If you want to suppress the 1st namespace, maybe manually issuing the `.. cpp:namespace: root` directive in the docs will solve it (don't think so though). If not, I'd recommend the `#ifndef` approach because we can't assume that generally. At most we could have some option that peels the 1st namespace off of everything, but I would still not make that the default.

> From a user documentation perspective, the sphinx namespace do exactly what I want. When I am in a different namespace, I can reference to the object with `subcomponent::A`, while in the same namespace with `A`.

Sure, but that wouldn't be broken either way I think. I am worried about complicating the code and preventing certain other use cases though, but I can't stress enough that I need to test some of these things.

Two things I'm hoping will work flawlessly and intuitively as per my concept:

```rst
.. Hypothetically autogenerated from Hawkmoth with autostruct namespace::symbol
.. cpp:struct:: namespace::symbol

.. Manually written
:cpp:struct:`namespace::symbol`

.. cpp:namespace-push:: namespace

:cpp:struct:`symbol`

.. cpp:namespace-pop::
```

Hence why I'd not want Hawkmoth messing with namespace directives: if we do that and the user uses namespaces at all, there's potential to break everything. And remember: Sphinx's namespaces are not specifically about C++ namespaces.

Anyway, let me try some of these things. Either way your parser changes are much simpler then I remember them being back in January, so this is quite helpful either way I think.

stephanlachnit:
> Well, names aside, Clang needs to parse namespaces in C++, it's only natural they show up in the syntax tree. But that's a bit of a red herring. I'm not contesting those directives can be used in this way, I'm proposing that it's not the best / easiest way to do it.

I don't know, it doesn't really to me. There is not really a difference between a class/enum namespace and a "namespace" namespace, both in how you actually use the C++ and how the sphinx documentation of such objects look like.

For example
```rst
.. cpp:enum-class:: TestE

   Test enum E

   .. cpp:enumerator:: A

      enum member A

   .. cpp:enumerator:: B

      enum member B
```

Here, the enum values aren't declared as `cpp:enumerator:: TestE::A` either. So why should it be different for a namespace?

> Hmm, not sure what the issue is. It seems very straightforward and actually independent of whether we use the namespace directives or not (think .. cpp:autoclass:: project::subcomponent::B or maybe .. cpp:autoclass:: B if we can manually issue the namespace directives). Can you elaborate?

Yes! Take a look at [`_var_type_fixup`](https://github.com/jnikula/hawkmoth/blob/commented-cursor/src/hawkmoth/commentedcursor.py#L230), [`_type_definition_fixup`](https://github.com/jnikula/hawkmoth/blob/commented-cursor/src/hawkmoth/commentedcursor.py#L308) or [`_get_inheritance`](https://github.com/jnikula/hawkmoth/blob/commented-cursor/src/hawkmoth/commentedcursor.py#L341). In my first attempt I tried to teach these functions namespaces, but unfortunately it's not as easy as just slapping the namespace in there (which is quite easy).

I did manage to get somewhere with my first approach (the one you are suggesting), but it got really complicated when trying to code up inheritance, templates, etc.

Thinking about a project-wide namespace again, in the documentation of my project I don't won't the global namespace in the documentation, as that is basically just clutter. The problem is that removing this namespace now becomes a non-trivial problem if you have complex declarations (e.g. classes with inheritance).

> If you want to suppress the 1st namespace, maybe manually issuing the .. cpp:namespace: root directive in the docs will solve it (don't think so though). If not, I'd recommend the #ifndef approach because we can't assume that generally. At most we could have some option that peels the 1st namespace off of everything, but I would still not make that the default.

It's not about cutting the global namespace by default (which I think would be a bad default anyway), but about having the option.

> Hence why I'd not want Hawkmoth messing with namespace directives: if we do that and the user uses namespaces at all, there's potential to break everything. And remember: Sphinx's namespaces are not specifically about C++ namespaces.

I don't see how that can happen tbh. You can still defines your own namespaces around these, I don't see a difference.

BrunoMSantos:
> For example
> [...]
> Here, the enum values aren't declared as `cpp:enumerator:: TestE::A` either. So why should it be different for a namespace?

Note those are scoped (indented) _inside_ the other directive, that's why there are no 'namespaces' there. I believe we could use Sphinx namepsaces to unscope them, but I don't see a use case for that in Hawkmoth.

> [...]
> I did manage to get somewhere with my first approach (the one you are suggesting), but it got really complicated when trying to code up inheritance, templates, etc.

I believe you, that is why I abandoned this until we refactored the parser to make these things easier (see below). Good news is that that refactor may be a lot closer now than ever :)

> [...] 
> It's not about cutting the global namespace by default (which I think would be a bad default anyway), but about having the option.

Yeah, I can get behind an explicit option if there's no better way (quite likely I think).

> > Hence why I'd not want Hawkmoth messing with namespace directives: if we do that and the user uses namespaces at all, there's potential to break everything. And remember: Sphinx's namespaces are not specifically about C++ namespaces.
> 
> I don't see how that can happen tbh. You can still defines your own namespaces around these, I don't see a difference.

I'll have to try some of these things no doubt, but I think I still haven't properly conveyed my main point.

To solve the individual symbol case, we need to do the difficult bits already: keep track of namespaces 'per symbol'. How else do we know in which namespace a symbol is when we're not simply dumping things in order from a source file? I.e. I still believe namespaces cannot be treated like the other cursor types or we will never support the filtered directives.

To my mind, we need to build the namespace stack somehow and pass it along to docstring 'per symbol', and that will be _a lot_ easier once we pass our 'cursor' to docstring instead of the current messy collection of strings. Once we have that, it's a matter of either appending it to the symbol names or issuing namespace directives depending on what works best, but that will be relatively simple either way. Currently I think the former is the best option for reasons I tried to illustrate as well, but that is not the pain point at the moment. The problem really is that I don't think this PR is compatible with (or not without some serious workarounds versus what I have in mind):

>  * [ ] Ensure that namespace objects are kept when filtering for specific symbol and not the entire file

jnikula:
Side note, while I haven't looked into this much, and the C++ support has pretty much been @BrunoMSantos 's thing, I've always regarded the `namespace` directives as something the *user* would use. That is, hawkmoth would generate the C++ domain directives with the full namespacing from the source in the names, and the user could - if they so choose - make it easier to reference stuff by using the `namespace` directives. If they get added automatically and behind the scenes by hawkmoth, I think it gets difficult for the user to use them.

stephanlachnit:
> To solve the individual symbol case, we need to do the difficult bits already: keep track of namespaces 'per symbol'. How else do we know in which namespace a symbol is when we're not simply dumping things in order from a source file? I.e. I still believe namespaces cannot be treated like the other cursor types or we will never support the filtered directives.

I don't have time to work on this for the weekend, but I don't think filtering is an unsolvable issue.

> To solve the individual symbol case, we need to do the difficult bits already: keep track of namespaces 'per symbol'. How else do we know in which namespace a symbol is when we're not simply dumping things in order from a source file? I.e. I still believe namespaces cannot be treated like the other cursor types or we will never support the filtered directives.
>
> To my mind, we need to build the namespace stack somehow and pass it along to docstring 'per symbol', and that will be a lot easier once we pass our 'cursor' to docstring instead of the current messy collection of strings. Once we have that, it's a matter of either appending it to the symbol names or issuing namespace directives depending on what works best, but that will be relatively simple either way. Currently I think the former is the best option for reasons I tried to illustrate as well, but that is not the pain point at the moment

Again, I tried, but it sounds a lot easier than it actually is. If clang has no magic "give me the full namespace of that object"-function (it probably does, I don't know), then resolving symbols is much harder than just pushing and poping from a stack.

I'm not opposed to this, it was my first attempt after all. What I do think however is that merging a solution like this is significantly better than entirely ignoring all C++ namespaces. The code changes are also non-intrusive, the classes in `docstring` can easily be removed and the `cursor.kind == CursorKind.NAMESPACE` check will stay in a similar form anyway.

> Side note, while I haven't looked into this much, and the C++ support has pretty much been @BrunoMSantos 's thing, I've always regarded the `namespace` directives as something the _user_ would use. That is, hawkmoth would generate the C++ domain directives with the full namespacing from the source in the names, and the user could - if they so choose - make it easier to reference stuff by using the `namespace` directives. If they get added automatically and behind the scenes by hawkmoth, I think it gets difficult for the user to use them.

I'm not really sure what is supposed to be difficult. I tested the implementation, and it works exactly as one would except as a user (see `CHIRP/Manager.hpp` in https://github.com/stephanlachnit/constellation-chirp-code/pull/2)

BrunoMSantos:
> I don't have time to work on this for the weekend, but I don't think filtering is an unsolvable issue.

I may be wrong of course, but I'd want some proof of concept to overcome my current strong belief that it doesn't work. I can also be flexible about merging it as a stop gap approach, but I also think this opens up a can of worms you're not fully considering despite it working in the code you tried it on so far.

> Again, I tried, but it sounds a lot easier than it actually is. If clang has no magic "give me the full namespace of that object"-function (it probably does, I don't know), then resolving symbols is much harder than just pushing and poping from a stack.

Note that using namespace directives don't magically solve anything, on the contrary. Consider this:

```c++
#include <string>
namespace a_namespace {
class a_class : public std::string {};
}
```

Is `std::string` inside `a_namespace`? If we 'push' the namespace, it would be inside in the documentation because Sphinx is not compiling any code obviously and so it cannot resolve namespaces in the same way as the compiler even if it tried.

So yeah, I'm hoping we can get Clang to lift some weight here indeed because we're a bit lost otherwise, but I haven't had the time to get there yet. I will be looking into it over the coming days though.

My current bet is that the easy parts are just as easy either way (honestly don't see the issues you're seeing with the alternative), the difficult ones will be easier without the directives I think (to which I think I gave convincing examples).

stephanlachnit:
@BrunoMSantos btw there is https://github.com/jbms/sphinx-immaterial/tree/main/sphinx_immaterial/apidoc/cpp which has proper support for C++ namespaces

stephanlachnit:
I think I found the solution for what I have been looking for: using `cursor.semantic_parent` gives either a) the parent namespace or b) the translation unit if there is none. Thus, we can just recursively look for semantic parents until we get the translation unit. The cool thing is that even work with classes that are defined inside classes.

stephanlachnit:
> I think I found the solution for what I have been looking for: using `cursor.semantic_parent` gives either a) the parent namespace or b) the translation unit if there is none. Thus, we can just recursively look for semantic parents until we get the translation unit. The cool thing is that even work with classes that are defined inside classes.

So this turned out to work very nicely, the code is quite short. However, there are still some edge cases where it does not work, e.g. class inheritance.

stephanlachnit:
Ah, unfortunately there are some annoying cases, e.g.
```cpp
namespace A {

    /** class B */
    template<class T> class B : public T {}; 

}
```
Gives:
```rst
.. cpp:class:: A::template<class T> A::B: public T

   class B
```
There are two ways to fix this:
- either make the namespace part of `name` and `decl_name`, or
- move template prefixes out of the name similar to function qualifiers

Opinions?

stephanlachnit:
> There are two ways to fix this:
> 
>     * either make the namespace part of `name` and `decl_name`, or
> 
>     * move template prefixes out of the name similar to function qualifiers
> 
> 
> Opinions?

I know went ahead with option number 1, as this is easier to implement. However, there is still one case that is not fixed, namely function arguments in `_var_type_fixup`. The clang interface here doesn't seem to be as straight forward (e.g. if you have a L-Value Reference), so I'll have to do some digging.

jnikula:
@stephanlachnit Hey, I'm sorry I haven't had the time to look into this. Not forgotten, I promise! Next year for sure. ;)

BrunoMSantos:
Hey, nice job figuring out the `semantic_parent` thing. That is indeed the kind of thing we would want from clang.

> There are two ways to fix this:
> 
>     * either make the namespace part of `name` and `decl_name`, or
>     * move template prefixes out of the name similar to function qualifiers
> 
> Opinions?

Sorry for the late reply, haven't had much time over the holidays.

`decl_name` is a ugly hack that we used to overcome the fact that parser and docstring were so tightly coupled and its meaning changes (changed?) with context. With the new cursor interface, we want to get rid of it and have much more contextual composition of the output in docstring. In particular to this problem, my hope is that docstring will use the properties `name`, `type`, `namespace`, `access_specifier`, etc. directly according to the specific cursor type.

So your 2nd option is the long term solution. Whether a short term, simpler one exists I don't know, but I would welcome it. Especially as I haven't even started with that part of the refactoring yet (it's coming! I swear :sweat_smile:).

> I know went ahead with option number 1, as this is easier to implement. However, there is still one case that is not fixed, namely function arguments in _var_type_fixup. The clang interface here doesn't seem to be as straight forward (e.g. if you have a L-Value Reference), so I'll have to do some digging.

That's good stuff! Keep us posted if you find out more ;)

Happy new year by the way!

jnikula:
This could use some tests. They also help understand what's being added functionally.

I think the changes look fine, apart from the nitpicks and tests. But I'd like a nod from @BrunoMSantos too.

BrunoMSantos:
> I think the changes look fine, apart from the nitpicks and tests. But I'd like a nod from @BrunoMSantos too.

I guess I could have been clearer, but I also do like the changes! This new solution addresses the only _real_ contention point with the original (the namespace push/pull approach).

But I'm also under the impression that there is still at least one known broken case in the current implementation (function arguments). And yeah, I would add quite a few tests before merging as well.

stephanlachnit:
Once I have time again (master colloquium coming up next week) I will add some tests and describe in detail what does and does not work. I think we can then merge this as a first step, and tackle function arguments in a second PR. I already played around with it and got some cases to work, but I think it takes a bit more work to get this fully working.

jnikula:
@stephanlachnit Sorry we've dropped the ball here a bit. :disappointed: 

Besides a nitpick, I think the code looks neat. But we could really use some test cases here to see what it does.

@BrunoMSantos Would you have a moment to also look at the changes, please? I'd really like to get the C++ support out of the limbo state of being there but not really very useful.

BrunoMSantos:
> @BrunoMSantos Would you have a moment to also look at the changes, please?

I did have a look at the time, but I took it to be a small interim improvement. I believe my last comment still stands?

> [...] But I'm also under the impression that there is still at least one known broken case in the current implementation (function arguments). And yeah, I would add quite a few tests before merging as well.

Otherwise this is great. I could also pick it up from here if @stephanlachnit prefers, but be warned: I have limited time and another personal project ramping up with a hard deadline :sweat_smile: 

> I'd really like to get the C++ support out of the limbo state of being there but not really very useful.

Indeed. And this PR would be a _huge_ step in the right direction.

stephanlachnit:
Sorry for not adding unit tests - unfortunately we had to focus more on the project and didn't have time to spend more time on hawkmoth, so we moved to breathe+doxygen for now :/ I really hope we can switch back one day, but for the foreseeable future I don't plan to work on this PR... so feel free to pick it up and add unit tests (except for known issues with template as function arguments the code is ready).

BrunoMSantos:
No worries of course, I think we all know where this is at. I'll pick it up then and thanks for the head start!

jnikula:
Variables don't get the namespace prefix:

```
namespace foo {
	/**
	 * foo_class
	 */
	class foo_class {
		/** member */
		int m;
	};

	/**
	 * Const.
	 */
	const int GLOBAL = 5;
};
```
Results in:
```
.. cpp:class:: foo::foo_class

   foo_class


   .. cpp:member:: private int m

      member


.. cpp:var:: const int GLOBAL

   Const.


```

jnikula:
Closing in favour of #247.

PR #212, opened 2023-11-09
Use -xc++-header instead of -xc++ in cpp domain

### LINKED ISSUE #208: C++: hawkmoth warns about `#pragma once in main file`
hawkmoth (or more likely, clang) warns about `#pragma once in main file`.

To reproduce, try to add documentation for this file:
```cpp
#pragma once

/*
 * Test doc
 */
void test_function() {}
```

The warning has no impact on the output.

stephanlachnit:
Closes #208

By default, hawkmoth added the -xc++ compiler argument to clang, which caused issues with pragma only declaration (See https://github.com/jnikula/hawkmoth/issues/208). Changig this to -xc++-header solves this issue, while retaining compatibility with compiled units.

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -53,9 +53,9 @@ class _AutoBaseDirective(SphinxDirective):
                             location=(self.env.docname, self.lineno))
 
     def __get_clang_args(self):
-        # Always pass `-xc++` to the compiler on 'cpp' domain as the first
+        # Always pass `-xc++-header` to the compiler on 'cpp' domain as the first
         # option so that the user can override it.
-        clang_args = ['-xc++'] if self._domain == 'cpp' else []
+        clang_args = ['-xc++-header'] if self._domain == 'cpp' else []
 
         clang_args.extend(self.env.config.hawkmoth_clang.copy())
 


jnikula:
Please make sure `make test` passes after making changes.

Please also write proper commit messages explaining why the change is being made.

Simply dropping the `-xc++` option won't cut it. The domain should dictate the language. It may not be best practise, but it should be possible to parse a `.h` file as C++. The test suite also parses `.c` files as C++, based on the domain.

BrunoMSantos:
This is weird. This should only have a positive impact, especially if your code is C++.

That was added to force C++ parsing because lots of people use `.h` in headers which by default will make Clang parse the file as C code. I'm quite confident the default is correct, and the user can always override it by passing `-xc`.

Disclaimer: I haven't followed the linked issue in detail.

jnikula:
> That was added to force C++ parsing because lots of people use `.h` in headers which by default will make Clang parse the file as C code. I'm quite confident the default is correct, and the user can always override it by passing `-xc`.

Passing `-xc++` to parse C++ headers works in most cases, but there are a few cases (e.g. `#pragma once`) where libclang makes a distinction on whether the primary file is a header or not. `-xc++-header` should be used for C++ headers instead.

I'm thinking maybe we should always add one of the `-xc++`, `-xc++-header`, `-xc`, `-xc-header` options at the parser level.

I'm not quite sure why yet, but I was also seeing the problem in the test suite when adding a `.h` file in the C domain. But only via the parser test, not via cli or sphinx tests. Curious.

stephanlachnit:
Tested now local, using `-xc++-header` instead of `-xc++` seems to solve the issue, compiled units still work.

jnikula:
Thanks for the report, the debugging, and the pull request.

However, I opted to go a bit further with this to better unify everything: #213

stephanlachnit:
Seems reasonable

PR #213, opened 2023-11-09
Unify -x<language> option usage

### LINKED ISSUE #208: C++: hawkmoth warns about `#pragma once in main file`

jnikula:
Move the clang language selection at a lower level, do it unconditionally, and add -header too. This unifies -x option across the extension, tests ,and cli.

Fixes #208

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -69,8 +69,8 @@ class _AutoBaseDirective(SphinxDirective):
         # Cached parse results per rst document
         parsed_files = self.env.temp_data.setdefault('hawkmoth_parsed_files', {})
 
-        # The output depends on clang args
-        key = (filename, tuple(clang_args))
+        # The output depends on domain and clang args
+        key = (filename, self._domain, tuple(clang_args))
 
         if key in parsed_files:
             return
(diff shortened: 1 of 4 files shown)


BrunoMSantos:
Looks great!

jnikula:
> Looks great!

Thanks; I additionally pushed some tests that would fail without the fixes in the series. Would you mind eyeballing them too, please?

test/c/pragma-once.h
```diff
@@ -0,0 +1,2 @@
+/** Ensure this file is parsed as a header */
+#pragma once
```
BrunoMSantos, line 2: Honestly, the whole error this would generate otherwise is a bit bogus if all it takes is to have `-xc*` or `-xc*-header`. Not sure we can rely on these tests failing long term is what I'm saying, so I'm not sure the tests are useful at all.

Not to mention it's a non standard pragma...

I understand why you're adding them and I won't oppose, but if it were only up to me, I wouldn't include these.

jnikula: Fair points. I was hoping to add some tests for the fixes, but this is probably a bit fragile indeed. Threw them away.

PR #215, opened 2023-11-14
Process docstring cli testing

jnikula:


src/hawkmoth/__main__.py
```diff
@@ -45,11 +57,20 @@ def main():
...
+                        ],
+                        help='Process docstring.')
+    compat.add_argument('--compat',
+                        choices=[
+                            'none',
+                            'javadoc-basic',
+                            'javadoc-liberal',
+                            'kernel-doc'
+                        ],
                         help='Compatibility options. See cautodoc_compat.')
```
BrunoMSantos, line 74: Consider raising `DeprecationWarning` :man_shrugging:

BrunoMSantos (approved):
Looks good ;)

src/hawkmoth/__main__.py
BrunoMSantos: Side note: we may also want to institutionalize a deprecation policy so we have a regular, well advertised schedule for dropping things. We have an important user now and everything <.<

jnikula: Yeah, the deprecation has been haphazard so far. If it's release based, I guess it'll have to be major releases. And we're not even at 1.0. :grin:

PR #218, opened 2023-11-18
Fix references to typedefs of anonymous struct/union/enum on clang 15 and earlier

jnikula:
Fix for #201


BrunoMSantos (approved):
Looks good. Small typo in the 1st commit message though: "[...],but should also cleaned up later."

jnikula:
> Looks good. Small typo in the 1st commit message though: "[...],but should also cleaned up later."

Thanks, fixed & merged!

PR #219, opened 2023-11-18
Add tips & tricks documentation

jnikula:
Add some tips for using Sphinx and Hawkmoth for C/C++ API documentation, etc.

### DIFF
--- a/doc/tips.rst
+++ b/doc/tips.rst
@@ -59,3 +59,16 @@ The Sphinx alternative is :external+sphinx:rst:dir:`literalinclude`. The
 .. _@include: https://www.doxygen.nl/manual/commands.html#cmdinclude
 
 .. _@snippet: https://www.doxygen.nl/manual/commands.html#cmdsnippet
+
+Using sphinx-autobuild with Hawkmoth
+------------------------------------
+
+`sphinx-autobuild`_ is a handy tool to automatically rebuild Sphinx
+documentation when you modify the ``.rst`` files, with live-reload in the
+browser.
+
+It's possible to have it auto rebuild and live-reload source documentation on
+source code changes by adding ``--watch <source root>`` option to
+``sphinx-autobuild``, where ``<source root>`` matches :py:data:`hawkmoth_root`.
+
+.. _sphinx-autobuild: https://github.com/executablebooks/sphinx-autobuild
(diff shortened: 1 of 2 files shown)


doc/tips.rst
```diff
@@ -59,3 +59,21 @@ The Sphinx alternative is :external+sphinx:rst:dir:`literalinclude`. The
 .. _@include: https://www.doxygen.nl/manual/commands.html#cmdinclude
 
 .. _@snippet: https://www.doxygen.nl/manual/commands.html#cmdsnippet
+
+Dealing With Generated Headers
+------------------------------
```
BrunoMSantos, line 64: This one is a bit much in my opinion, but I guess it's fine? I do see this as a workaround for a problem that is entirely unrelated to hawkmoth. If it were me though, I'd make the documentation depend on generating the headers, let the build system do what it's meant to do. :man_shrugging:

BrunoMSantos (approved):
Cool idea and nice set of tips. Apart from the one I objected to, but I'm ok with it really. I'll leave it up to you ;)

doc/tips.rst
jnikula: FWIW this is the approach taken in Mesa. I think it's a tall order to make the documentation build depend on generating the headers, which in turn depend on additional tooling and whatnot.

The reason this is related to Hawkmoth is that we use the compiler to extract the documentation, and therefore hit the the problem of missing headers. AFAIK Doxygen uses a parser of its own, and doesn't care if it can't find headers.

jnikula: Arguably it's a problem in Mesa that the Sphinx build is not part of the Meson build files, at all. The documentation is generated simply by running `sphinx-build` in CI. But that's actually the same as on Read the Docs.

BrunoMSantos: So, I'm still ok with this. But maybe it wouldn't trigger me as much if it were a more nuanced tip... One that points out that the files must be created somehow and that it should typically be handled by the build system and possibly with `conf.py`. And also that the code doesn't have to be fully compilable as we skip the function bodies, so the generated files can be stubs instead.

As it stands, I think this particular advise is a bit out of place with Hawkmoth's ethos if not bad general advise altogether. Using a compiler underneath is _the_ defining feature of hawkmoth, it stands to reason the code must be compilable.

If I were to advise on this topic, I would rather recommend 'use a build system, take the flags from there and ...':
- If performance is not an issue, generate the files.
- If performance is an issue, generate the smallest adequate stub from the same sources / tools (single source of truth and whatnot).

Also, if dependent on calling `sphinx-build` without build system, I'd still call the build system from the `conf.py` to generate those files.

jnikula: Fair enough. I think I'll leave the section out altogether, at least for now.

jnikula:
And merged. Thanks for your input @BrunoMSantos !

PR #220, opened 2023-11-18
doccursor: expose cursor location

BrunoMSantos:
This is used to get the file name and line for error messages in the parser. Naturally we need to expose it.


stephanlachnit (approved):
LGTM

PR #221, opened 2023-11-19
docstring: rename _nest method

stephanlachnit:
Currently, the `Docstring` class has a `_nest` member variable for the current nesting level, and a `_nest` method to nest some lines. This commit changes the `_nest` method to `_nest_lines` to avoid duplicate symbols.

### DIFF
--- a/src/hawkmoth/docstring.py
+++ b/src/hawkmoth/docstring.py
@@ -126,7 +126,7 @@ class Docstring():
         return line_offset
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


jnikula:
That's indeed a bit confusing, thanks!

jnikula:
Oh, already merged, and it's not documented anywhere, but for future reference I'd prefer the commit messages to be word wrapped to something like 72 chars.

PR #222, opened 2023-11-19
setup: move flake8 filename patterns to setup.cfg

jnikula:
Commit dc17f4165508 ("setup: move flake8 configuration to setup.cfg") moved flake8 ignore patterns and line length config to setup.cfg. Also move the filename patterns there.

### DIFF
--- a/Makefile.local
+++ b/Makefile.local
@@ -20,7 +20,7 @@ RST := $(RST) CHANGELOG.rst CONTRIBUTING.rst README.rst
 # Static analysis
 .PHONY: check
 check:
-	flake8 src/hawkmoth test
+	flake8
 
 .PHONY: check-rst
 check-rst:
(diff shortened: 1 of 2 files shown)


BrunoMSantos (approved):
Well noted, but actually, seems like flake 8 already finds everything on its own, this is only needed if we don't want to apply it somewhere as an exception. Maybe we want to remove it altogether?

Otherwise it's fine.

jnikula:
> Well noted, but actually, seems like flake 8 already finds everything on its own, this is only needed if we don't want to apply it somewhere as an exception. Maybe we want to remove it altogether?

Then we'd need to add excludes, because it finds absolutely everything in the virtual environment, checks `doc/conf.py`, etc. I prefer it this way.

> 
> Otherwise it's fine.

BrunoMSantos:
Oh, ok. I personally like checking `conf.py`. Oh well, I'll fight that one some other day :P

PR #224, opened 2023-11-19
examples: mention that source links are optional

jnikula:
Some folks might be put off by the source links, others might struggle figuring out how they're enabled. Pointing at the option helps both.


BrunoMSantos (approved):
Looks good ;)

PR #228, opened 2023-11-19
Revert "setup: move flake8 filename patterns to setup.cfg"

jnikula:
This reverts commit 9c20afb4702c31050b956af7612614a3adffde9b.

Turns out the filename config doesn't do what I thought it does. It lists filename patterns within paths, not files or directories to check, and the change skipped everything.


jnikula:
Decided to just merge this one.

PR #229, opened 2023-11-25
parser: remove useless if

stephanlachnit:
No need to first check if the kind is in a list and then check which of the two items it is.


jnikula:
Thanks! Leftovers from (or, made possible by!) the refactors.

PR #230, opened 2023-11-26
style: follow pep8 on lambda assignments

jnikula:
PEP8 https://peps.python.org/pep-0008/ says:

Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier.

Fix them, and drop the flake8 exception.


BrunoMSantos:
Looks great and I learnt something today ;)

PR #232, opened 2024-01-11
build: bump python dependency to 3.8

jnikula:
Python 3.6 and 3.7 have reached end of life [1].

[1] https://devguide.python.org/versions/


BrunoMSantos (approved):
All good!

PR #234, opened 2024-01-27
Makefile updates

jnikula:
Long overdue cleanup of some old Makefile cruft.

First lift the two trivial and non-controversial commits from #231, and continue from there.


src/hawkmoth/Makefile.local
```diff
@@ -1,5 +1,5 @@
 # -*- makefile -*-
 
-dir := src/hawkmoth
+src_dir := src/hawkmoth
```
BrunoMSantos, line 3: Being overly pedantic here, but this would fit better with the 1st commit. No biggie though.

Makefile.rules
```diff
@@ -15,9 +15,6 @@ endif
 # Otherwise, print the full command line.
 quiet ?= $($1)
 
-%.html: %.rst
-	rst2html --strict --no-raw $< > $@
```
BrunoMSantos, line 19: This is being added back in the following commit. I understand the rationale, so if you want this in the history for future reference it's fine with me, just wanted to make sure it wasn't a mistake.

Makefile
```diff
@@ -1,5 +1,7 @@
+# Unconditionally make all targets; they're all phony
+MAKEFLAGS += --always-make
```
BrunoMSantos, line 2: Hmm, while I appreciate your reasons for it, I'd rather add `PHONY` where it's missing. These things don't even change that often and are not too disruptive if we do find an issue every now and then.

That said, it's ok with me if you really prefer it this way.

BrunoMSantos (approved):
Looks good to me despite the non 0 comments. Merge at will!

src/hawkmoth/Makefile.local
jnikula: I guess I made this drive-by change, and that inspired me to do the `docdir` change.

Makefile.rules
jnikula: Confused. Where is this being added back?

BrunoMSantos: Oh man... I _think_ what happened was that I clicked on the referenced commits to have a look and then proceeded to consider one of them part of the PR :man_facepalming:

Ignore me -.-

Makefile
jnikula: Yeah, it's just that we don't really use make to *build* anything. It's only for running a bunch of recipes, really. So everything is "phony". If I could get away with switching to something like https://github.com/casey/just, which is basically make but without any of the build stuff and all the historical baggage, I'd consider it.

Makefile.rules
jnikula: Happens to the best of us! :rofl:

Makefile
jnikula: The alternative to adding `.PHONY: foo` for each target `foo` is to make each target depend on a `FORCE` target. See https://www.gnu.org/software/make/manual/make.html#Force-Targets

So instead of

```
.PHONY: clean
clean:
        ...
```
We'd have

```
clean: FORCE
        ...
```

That's a bit easier on the eye, perhaps?

BrunoMSantos: Well, we could end up creating rules to build packages or to do code generation... Things change. And then we'll need to add these back after someone notices that things are being regenerated when they shouldn't, which is much more puzzling to debug by the way.

`always-make` would be the last assumption I would make as to why something was being regenerated when I didn't want it to. Especially as it's so far removed from the rule definition itself. In fact, I'd only use that switch explicitly and for debugging purposes. Maybe. Conversely, something doesn't run, I know where to look.

Anyway, it really is up to you.

jnikula: Fair enough. I assume you wrote that before my suggestion about `FORCE` - compromise?

BrunoMSantos: > The alternative to adding `.PHONY: foo` for each target `foo` is to make each target depend on a `FORCE` target. See https://www.gnu.org/software/make/manual/make.html#Force-Targets

From the manual: `Using ‘.PHONY’ is more explicit and more efficient.` :sweat_smile: 

Seriously, if you don't like make, we should be considering meson or something like that. But as long as we're in Rome...

BrunoMSantos: > Fair enough. I assume you wrote that before my suggestion about `FORCE` - compromise?

I did. And the following one before I saw this one too. I don't think it's a good compromise by the way. This is the worst option so far: doesn't solve your issue and is equally bad from my perspective.

jnikula: Okay, okay! :smile: 

Dropped this, updated .PHONY deps, and split out the src_dir change to a separate commit.

PR #237, opened 2024-03-24
parser: fix linkage spec detection

### LINKED ISSUE #235: 0.17.0: pytest fails in 4 units: cpp mixed-linkage
I'm packaging your module as an rpm package so I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix> using `installer` module
- run pytest with $PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>
- build is performed in env which is *`cut off from access to the public network`* (pytest is executed with `-m "not network"`)

<details>
<summary>Here is pytest output:</summary>

```console
Package                       Version
----------------------------- -----------
alabaster                     0.7.16
Babel                         2.14.0
build                         1.1.1
charset-normalizer            3.3.2
clang                         18.1.2
distro                        1.9.0
docutils                      0.20.1
exceptiongroup                1.1.3
gpg                           1.23.2
idna                          3.6
imagesize                     1.4.1
importlib_metadata            7.0.1
iniconfig                     2.0.0
installer                     0.7.0
Jinja2                        3.1.3
libdnf                        0.73.0
MarkupSafe                    2.1.3
packaging                     24.0
pluggy                        1.4.0
Pygments                      2.17.2
pyproject_hooks               1.0.0
pytest                        8.1.1
python-dateutil               2.9.0.post0
requests                      2.31.0
setuptools                    69.1.1
snowballstemmer               2.2.0
Sphinx                        7.2.6
sphinxcontrib-applehelp       1.0.8
sphinxcontrib-devhelp         1.0.5
sphinxcontrib-htmlhelp        2.0.5
sphinxcontrib-jsmath          1.0.1
sphinxcontrib-qthelp          1.0.7
sphinxcontrib-serializinghtml 1.1.10
strictyaml                    1.7.3
tokenize_rt                   5.2.0
tomli                         2.0.1
urllib3                       1.26.18
wheel                         0.43.0
zipp                          3.17.0
```
</details>

<details>
<summary>List of installed modules in build env:</summary>

```console
```
</details>

Please let me know if you need more details or want me to perform some diagnostics.

BrunoMSantos:
Clang 18 now issues the right cursor kind for `extern "C"` blocks, so we need to extend the detection code.

Fixes #235.

-->8--

This should reliably fix the detection of those blocks on all supported versions of Clang, but a few improvements might still be possible:

1. We should probably _not_ check for UNEXPOSED_DECL unless we are on pre-18 versions to keep it neater. This would require a version check on the libclang that is actually loaded.

2. It is unclear whether we can avoid token matching for v18+. In principle, if the cursor kind is LINKAGE_SPEC, the children cursors are in the C domain. But is that a guarantee or just a strong heuristic? If not, is there a way of checking without token matching?

3. We should come up with a way to systematically mark code portions that need to be revisited as we update minimum requirements and such to avoid dead code. I can only think of another instance where we might want to do something along those lines (anonymous symbol handling), but I think it may be good to think about this soon-ish rather than later just in case more cases pop up.

### DIFF
--- a/src/hawkmoth/doccursor.py
+++ b/src/hawkmoth/doccursor.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2016-2023 Jani Nikula <jani@nikula.org>
-# Copyright (c) 2018-2023 Bruno Santos <brunomanuelsantos@tecnico.ulisboa.pt>
+# Copyright (c) 2018-2024 Bruno Santos <brunomanuelsantos@tecnico.ulisboa.pt>
 # Licensed under the terms of BSD 2-Clause, see LICENSE for details.
 
 from clang.cindex import (
@@ -146,10 +146,11 @@ class DocCursor:
         """Get children cursors."""
         domain = self.domain
 
-        # Identify `extern "C"` blocks and change domain accordingly. For some
-        # reason, the Python bindings don't return the cursor kind LINKAGE_SPEC
-        # as one would expect, so we need to do it the hard way.
-        if domain == 'cpp' and self.kind == CursorKind.UNEXPOSED_DECL:
+        # Identify `extern "C"` blocks and change domain accordingly.
+        # Prior to Clang 18, the Python bindings don't return the cursor kind
+        # LINKAGE_SPEC as one would expect, so we need to do it the hard way.
+        if domain == 'cpp' and self.kind in [CursorKind.LINKAGE_SPEC,
+                                             CursorKind.UNEXPOSED_DECL]:
             tokens = self.get_tokens()
             ntoken = next(tokens, None)
             if ntoken and ntoken.spelling == 'extern':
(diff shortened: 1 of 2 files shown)


BrunoMSantos:
Point 3. might actually be made redundant by point 1.

jnikula (approved):

jnikula:
Replying to https://github.com/jnikula/hawkmoth/issues/235#issuecomment-2016783664

> I don't think any of those points are particularly interesting as long as it works reliably and I think it does, but I was letting you have a say on them I guess. Point 2 particularly, could be investigated a bit more (I spent maybe 15min on this, most of it setting up my environment to replicate the issue).

I don't think any of the points are blockers for merging this as-is now, but rather things to consider for the future.

There hasn't been much activity lately, but I might do a release sooner rather than later with the fixes.

jnikula:
> 1. This would require a version check on the libclang that is actually loaded.

I don't actually see a way to cleanly check the libclang version from python-clang. Do you have any ideas?

BrunoMSantos:
> I don't think any of the points are blockers for merging this as-is now, but rather things to consider for the future.

Very well, I'm leaning that way too.

> I don't actually see a way to cleanly check the libclang version from python-clang. Do you have any ideas?

Not really. Actually I asked you in IRC after having given up already.

Best I could come up with was to look into `cindex.Config.library_{file,path}` and infer it from there or, if `None`, look into the LD path and so on. Not clean at all...

jnikula:
Merged, thanks for the prompt fix @BrunoMSantos!

PR #238, opened 2024-03-24
Docker updates

jnikula:
When setting up an environment to test clang 18 in a container, I stumbled on some issues with the docker test stuff. Fix them.

@BrunoMSantos Not sure if you care much about reviewing these kind of changes.


BrunoMSantos:
I can always learn something at least ;)

And yeah, this looks good and useful! It would have saved me some time on the last PR.

PR #239, opened 2024-03-24
Macro named variable argument

jnikula:
Add support for the named variable argument in variadic macros i.e. `FOO(arg...)` style, and remove a long standing FIXME comment.

https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html


BrunoMSantos:
Hey, sorry for taking so long. I reviewed this over the weekend but then forgot to say something :sweat_smile: 

Looks good to me though!

jnikula:
> Hey, sorry for taking so long. I reviewed this over the weekend but then forgot to say something 😅
> 
> Looks good to me though!

Thanks, appreciated, as always!

PR #241, opened 2024-04-23
Start using hatch for Python project management

jnikula:
Hatch seems more user-friendly than setuptools and twine, and offer more stuff like environments (which could replace the `venv` thing in the future).

I don't think this should make a difference to contributors or users.


jnikula:
> Hey, sorry for taking so long. Been quite busy with other projects.

No worries at all!

> This looks ok to me. Though I can't help but wince at just how many different ways there are to do something like this 😅

I know!

> I can't say I know how this improves things either, but I'm happy to take your word for it. Frankly I have no horse in this race.

The main thing for me is reducing the number of tools for building/uploading, with fewer sharp edges, maybe using it to replace the `venv` script which I believe only I use anyway, and more. Mainly I'm picking hatch because it's a PyPA project.

> Part of me would like to have kept an historical record of what the clean setuptools version was in case we had to roll back for whatever reason, instead of changing tooling and resulting package in one go. It's too minor though, up to you.

Yeah, I also felt a bit like too much at once, but it's non-zero effort to convert to pyproject.toml with setuptools. Maybe I'll have another look at that, but if it's too much trouble, I'll just merge this.

> Need a small update to my packaging I guess, no impact beyond that.

What's that, out of curiosity?

BrunoMSantos:
> > Part of me would like to have kept an historical record of what the clean setuptools version was in case we had to roll back for whatever reason, instead of changing tooling and resulting package in one go. It's too minor though, up to you.
> 
> Yeah, I also felt a bit like too much at once, but it's non-zero effort to convert to pyproject.toml with setuptools. Maybe I'll have another look at that, but if it's too much trouble, I'll just merge this.

Yup, sounds good!

> > Need a small update to my packaging I guess, no impact beyond that.
> 
> What's that, out of curiosity?

[Here,](https://aur.archlinux.org/packages?O=0&K=hawkmoth) Nothing special... I know someone else has installed the package at some point because I got a complaint it was out of date once, but not too popular yet :sweat_smile:

jnikula:
> > Yeah, I also felt a bit like too much at once, but it's non-zero effort to convert to pyproject.toml with setuptools. Maybe I'll have another look at that, but if it's too much trouble, I'll just merge this.
> 
> Yup, sounds good!

Wasn't too bad actually, and I compared the build results after each commit.

> > > Need a small update to my packaging I guess, no impact beyond that.
> > 
> > What's that, out of curiosity?
> 
> [Here,](https://aur.archlinux.org/packages?O=0&K=hawkmoth) Nothing special... I know someone else has installed the package at some point because I got a complaint it was out of date once, but not too popular yet 😅

Heh, though I did mean what *changes* are required due to this?

BrunoMSantos:
> Heh, though I did mean what _changes_ are required due to this?

Oh, don't know. But I expect I need to tweak the PKGBUILD a bit. So (e.g.) [in the official release package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-sphinx-hawkmoth) change the build and install lines I guess.

jnikula:
> > Heh, though I did mean what _changes_ are required due to this?
> 
> Oh, don't know. But I expect I need to tweak the PKGBUILD a bit. So (e.g.) [in the official release package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-sphinx-hawkmoth) change the build and install lines I guess.

I think it should actually work as-is. 

```
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

tells `python3 -m build` to install and use hatchling as the build backend.

BrunoMSantos:
Hmm, even better then! I saw `hatch build` in the makefile and assumed that would be the right way(tm) to do it. I still need to add hatch to the build dependencies anyway, so not totally 0 changes.

jnikula:
@BrunoMSantos Pushed, please let me know if you hit any issues with packaging!

jnikula:
@BrunoMSantos Oh, might as well ask here about one thing. `setup.cfg` is now empty save for `flake8` config, because `flake8` doesn't support configuration via `pyproject.toml`, and that's a whole can of worms (https://github.com/PyCQA/flake8/issues/234 and they're now closing all PRs/issues even mentioning `pyproject.toml` as spam...).

What to do?

- Leave it as it is.
- Rename `setup.cfg` to `.flake8`.
- Install https://github.com/john-hen/Flake8-pyproject `flake8` plugin to read config from `pyproject.toml`.

BrunoMSantos:
> @BrunoMSantos Oh, might as well ask here about one thing. `setup.cfg` is now empty save for `flake8` config, because `flake8` doesn't support configuration via `pyproject.toml`, and that's a whole can of worms ([PyCQA/flake8#234](https://github.com/PyCQA/flake8/issues/234) and they're now closing all PRs/issues even mentioning `pyproject.toml` as spam...).

Had to read up on this a bit.

> What to do?
>     * Leave it as it is.

Of the 3, I think this is the one. There might be other 3rd party tools that don't yet support the toml one and we might need to extend it before we get rid of it.

>     * Rename `setup.cfg` to `.flake8`.

I'm fine with this as a 2nd best option.

>     * Install https://github.com/john-hen/Flake8-pyproject `flake8` plugin to read config from `pyproject.toml`.

Please, no :sweat_smile: 


But here's another option just in case: use pyflakes, pycodestyle and mccabe directly. Pycodestyle supports the toml file and I don't think we will be configuring pyflakes or mccabe away from any default, but they may support the toml file too :man_shrugging: 

Honestly, I never use the flake8 wrapper in my own projects, so I'd switch in a heartbeat. I even had to [hack my configs](https://gitlab.com/brunomanuelsantos/dotfiles/-/blob/837d608035c804fc2b674c67fc7b824565c46897/files/.config/nvim/lua/my-plugins/plugins.lua#L438) for those few projects I work on that use flake8. And now you know xD

jnikula:
> But here's another option just in case: use pyflakes, pycodestyle and mccabe directly. Pycodestyle supports the toml file and I don't think we will be configuring pyflakes or mccabe away from any default, but they may support the toml file too 🤷‍♂️

I don't think pycodestyle supports pyproject.toml, though: https://pycodestyle.pycqa.org/en/latest/intro.html#configuration

Also has the same maintainer as flake8, so would be suprising if it did.

> Honestly, I never use the flake8 wrapper in my own projects, so I'd switch in a heartbeat. I even had to [hack my configs](https://gitlab.com/brunomanuelsantos/dotfiles/-/blob/837d608035c804fc2b674c67fc7b824565c46897/files/.config/nvim/lua/my-plugins/plugins.lua#L438) for those few projects I work on that use flake8. And now you know xD

:eyes:

BrunoMSantos:
> > But here's another option just in case: use pyflakes, pycodestyle and mccabe directly. Pycodestyle supports the toml file and I don't think we will be configuring pyflakes or mccabe away from any default, but they may support the toml file too 🤷‍♂️
> 
> I don't think pycodestyle supports pyproject.toml, though: https://pycodestyle.pycqa.org/en/latest/intro.html#configuration

Hmm, could have sworn it was supported. Maybe the project I saw it in was using some other tool to bridge the gap like the one you pointed out for flake8, maybe I just remember wrong. Sorry for the misdirection then :/

> Also has the same maintainer as flake8, so would be suprising if it did.

Can't say I blame him either way, I did mention the wincing :sweat_smile:

jnikula:
> Hmm, could have sworn it was supported. Maybe the project I saw it in was using some other tool to bridge the gap like the one you pointed out for flake8, maybe I just remember wrong. Sorry for the misdirection then :/

No problem. And thanks for the discussion. I think I'm just going to do nothing about this.

PR #246, opened 2024-08-31
Transformations deprecation

jnikula:
Time to start getting rid of `cautodoc_transformations`, `hawkmoth.ext.transformations`, and `hawkmoth.util.doccompat`. The better alternatives i.e. `hawkmoth.ext.javadoc` extension and `hawkmoth-process-docstring` event have been in place since v0.13.


jnikula:
Also remove the deprecated cautodoc_root and cautodoc_clang config options.

src/hawkmoth/__init__.py
```diff
@@ -360,19 +360,8 @@ def _deprecate(conf, old, new, default=None):
...
-    app.add_config_value(
-        'hawkmoth_root',
-        lambda conf: _deprecate(conf, 'cautodoc_root', 'hawkmoth_root', app.confdir),
-        'env', [str]
-    )
-    app.add_config_value(
-        'hawkmoth_clang',
-        lambda conf: _deprecate(conf, 'cautodoc_clang', 'hawkmoth_clang', []),
-        'env', [list]
-    )
```
BrunoMSantos, line 375: I would also delete `_deprecate` since it's no longer used. Sure it could be useful in the future, but it's so simple to add it back even if we need it in its exact form and currently it's dead code.

BrunoMSantos (approved):
Good riddance!

Have you been keeping tabs on Mesa? I don't think they'll be impacted from what I've seen, but you've been more involved.

src/hawkmoth/__init__.py
jnikula: I thought I did. But apparently I only thought, but didn't really. Will fix.

jnikula:
> Have you been keeping tabs on Mesa? I don't think they'll be impacted from what I've seen, but you've been more involved.

Double-checked, should not have an impact.

src/hawkmoth/__init__.py
BrunoMSantos: Looks good!

dvdhrm:
I have been using the transformations to get kerneldoc-style comments working with hawkmoth, and I have to admit I am a bit confused what the recommended alternative is.

So far I used:

```
def kerneldoc_converter(comment):
    ...comment = re.sub("...", comment)...

cautodoc_transformations = {
    'kerneldoc': kerneldoc_converter,
}
```

I assume I am meant to use `hawkmoth-process-docstring`, but I have not found a simple example how to hook this up, yet. It just keeps pointing further into sphinx documentation, and I could not link all the pieces together, yet.

Any help would be appreciated! And maybe I can provide a simple example for the documentation on how to achieve this.

PR #247, opened 2024-08-31
Basic C++ namespace support

jnikula:
This incorporates @stephanlachnit's work from #211 and adds some tests on top.

Let's be honest, there are still some gaps here. For example, variables within a namespace don't get the namespace prefix in documentation. However, I'm making the argument this is a net positive improvement over the status quo, regardless of those gaps. This is progress. This can be built upon.

The tests added merely check against the current output, not what might be the desired output. I don't even have enough recent experience on modern C++ to say what the desired prefixing would be.

Anyway, unless @stephanlachnit and @BrunoMSantos are vehemently opposed, I'm going to merge this as-is.

Let's just move things forward.


jnikula:
Also, fixes #209.

PR #248, opened 2024-08-31
Prep work for running test suite on macos-latest

jnikula:
GitHub allows running workflows on macos in addition to ubuntu. I'd potentially like to start running tests on it, but there are some hiccups. I'm not quite there yet with fixing everything in setting up the workflow, but here are two small fixes for starters.

EDIT: This is expanding quite a bit...

EDIT: This + the actual workflow is #249. I'm keeping these separate as this is fairly straightforward prep work, and the workflow doesn't pass.

EDIT: And back to basics. This is just the good stuff now.

### DIFF
--- a/test/test_cautodoc.py
+++ b/test/test_cautodoc.py
@@ -63,12 +63,12 @@ class ExtensionTestcase(testenv.Testcase):
         with tempfile.TemporaryDirectory() as srcdir:
             with open(os.path.join(srcdir, 'index.rst'), 'w') as f:
                 f.write(input_str)
-            return self._sphinx_build(srcdir)
+            return self._sphinx_build(os.path.realpath(srcdir))
 
     def _sphinx_build_file(self, input_filename):
         with tempfile.TemporaryDirectory() as srcdir:
             shutil.copyfile(input_filename, os.path.join(srcdir, 'index.rst'))
-            return self._sphinx_build(srcdir)
+            return self._sphinx_build(os.path.realpath(srcdir))
 
     def get_output(self):
         input_str = ''.join([d.get_directive_string() for d in self.directives])
--- a/test/test_parser.py
+++ b/test/test_parser.py
@@ -2,8 +2,6 @@
 # Copyright (c) 2018-2023, Jani Nikula <jani@nikula.org>
 # Licensed under the terms of BSD 2-Clause, see LICENSE for details.
 
-import os
-
 import pytest
 
 from hawkmoth import docstring
@@ -81,7 +79,7 @@ class ParserTestcase(testenv.Testcase):
             roots[key] = root
 
             for error in errors:
-                errors_str += f'{error.level.name}: {os.path.basename(error.filename)}:{error.line}: {error.message}\n'  # noqa: E501
+                errors_str += f'{error.level.name}: {error.get_message(basename=True)}\n'
 
         for directive in self.directives:
             filename = directive.get_input_filename()
--- a/Makefile
+++ b/Makefile
@@ -2,4 +2,4 @@
 .PHONY: default
 default: help
 
-include $(shell find -name Makefile.local)
+include $(shell find . -name Makefile.local)
--- a/src/hawkmoth/parser.py
+++ b/src/hawkmoth/parser.py
@@ -65,12 +65,14 @@ class ParserError:
     line: int
     message: str
 
-    def get_message(self):
+    def get_message(self, basename=False):
         if self.filename:
+            filename = os.path.basename(self.filename) if basename else self.filename
+
             if self.line is not None:
-                return f'{self.filename}:{self.line}: {self.message}'
+                return f'{filename}:{self.line}: {self.message}'
             else:
-                return f'{self.filename}: {self.message}'
+                return f'{filename}: {self.message}'
         else:
             return f'{self.message}'
 


jnikula:
I'm not sure if this is the right thing to do, though.

test/test_parser.py
```diff
@@ -81,7 +79,7 @@ def get_output(self):
             roots[key] = root
 
             for error in errors:
-                errors_str += f'{error.level.name}: {os.path.basename(error.filename)}:{error.line}: {error.message}\n'  # noqa: E501
+                errors_str += f'{error.level.name}: {error.get_message(basename=True)}\n'
```
BrunoMSantos, line 82: I'm a bit confused on this one, and bear in mind I'm going by the partial diffs and the commit messages only.

You mention a possibility of having filename being `None`, is that right? The `basename=True` path is not really about that though, it's about calling `basename` or not, but still using the original name, which apparently may be `None`. If it were about that, it could check for `None` without any additional argument like we do already with the `line`.

test/test_cautodoc.py
```diff
@@ -54,7 +54,7 @@ def _sphinx_build(self, srcdir):
             output_filename = os.path.join(app.outdir, f'index.{output_suffix}')
 
         # Remove paths from warning output for comparison
-        output_errors = re.sub(rf'(?m)^{srcdir}/index.rst:[0-9]+: ([^:]*: )(/[a-zA-Z0-9._-]+)*/',
+        output_errors = re.sub(rf'(?m)^.*{srcdir}/index.rst:[0-9]+: ([^:]*: )(/[a-zA-Z0-9._-]+)*/',
```
BrunoMSantos, line 57: Do you think it may be worth checking where the prefix comes from? I don't think it matters all that much, but it still feels a bit weird. Maybe we could also have the regex match only what we know may appear instead of filtering away anything, including (e.g.) some root directory added by mistake.

BrunoMSantos (commented):
Good goal and it's a tentative approve from me, but see my other comments 1st.

test/test_cautodoc.py
jnikula: So I didn't dig into it that much, but apparently `/private` is a historical thing on macOS, and e.g. `/tmp` is a symlink to `/private/tmp`. The discrepancy might come from some parts normalizing the symlink away and some parts not.

I guess an alternative might be doing `os.path.abspath()` on the paths we get from `tempfile.TemporaryDirectory()` so everything is handled in the same "domain" so to speak.

BrunoMSantos: I see, matching anything including root paths may not be so weird given that `/private` is indeed a path and not some other kind of prefix. Otherwise, yours might be a better alternative to mine indeed if you do feel like changing it. I may be overlooking other complications though.

test/test_parser.py
BrunoMSantos: Ah, I see. Actually it made sense now that I read it again. Yeah, feel free to disregard ;)

test/test_cautodoc.py
jnikula: This also fixes the issue, better?

```
diff --git a/test/test_cautodoc.py b/test/test_cautodoc.py
index 2c957d3d4b1f..178a6bf83290 100755
--- a/test/test_cautodoc.py
+++ b/test/test_cautodoc.py
@@ -63,12 +63,12 @@ class ExtensionTestcase(testenv.Testcase):
         with tempfile.TemporaryDirectory() as srcdir:
             with open(os.path.join(srcdir, 'index.rst'), 'w') as f:
                 f.write(input_str)
-            return self._sphinx_build(srcdir)
+            return self._sphinx_build(os.path.realpath(srcdir))
 
     def _sphinx_build_file(self, input_filename):
         with tempfile.TemporaryDirectory() as srcdir:
             shutil.copyfile(input_filename, os.path.join(srcdir, 'index.rst'))
-            return self._sphinx_build(srcdir)
+            return self._sphinx_build(os.path.realpath(srcdir))
 
     def get_output(self):
         input_str = ''.join([d.get_directive_string() for d in self.directives])

```

jnikula: abspath does *not* fix the issue, it's realpath that's needed.

jnikula:
Switched to using `os.path.realpath(srcdir)` to handle the symlinks, which is a more general approach I think. Also improved the commit messages around `get_message()` use slightly.

Testing the macos workflows in a private repo, I'm getting pretty close to passing the tests, but not quite.

jnikula:
With the xcode (or whatever the "stock" toolchain is called) version of libclang, I'm hitting e.g.:

```
E         + WARNING: using-alias.cpp:2: alias declarations are a C++11 extension
E         + WARNING: using-alias.cpp:5: alias declarations are a C++11 extension
E         + WARNING: using-alias.cpp:9: alias declarations are a C++11 extension
E         + WARNING: using-alias.cpp:12: variadic templates are a C++11 extension
E         + WARNING: using-alias.cpp:13: alias declarations are a C++11 extension
E         + WARNING: using-alias.cpp:16: variadic templates are a C++11 extension
E         + WARNING: using-alias.cpp:17: alias declarations are a C++11 extension
```

and

```
E         + ERROR: struct-class.cpp:26: unknown type name 'constexpr'
E         + /Users/runner/work/github-actions-playground/github-actions-playground/test/cpp/struct-class.cpp:23: WARNING: Error in declarator or parameters-and-qualifiers
E         + If pointer to member declarator:
E         +   Invalid C++ declaration: Expected identifier in nested name. [error at 48]
E         +     public static constexpr  static_constexpr_member
E         +     ------------------------------------------------^
E         + If declarator-id:
E         +   Invalid C++ declaration: Expected identifier in nested name. [error at 48]
E         +     public static constexpr  static_constexpr_member
E         +     ------------------------------------------------^
E         +
```

and

```
E         + ERROR: class.cpp:28: unknown type name 'constexpr'
E         + ERROR: class.cpp:34: expected ';' at end of declaration list
E         + ERROR: class.cpp:37: unknown type name 'constexpr'
E         + WARNING: class.cpp:70: deleted function definitions are a C++11 extension
E         + WARNING: class.cpp:73: 'override' keyword is a C++11 extension
E         + WARNING: class.cpp:91: defaulted function definitions are a C++11 extension
```

and

```
E         + ERROR: namespace.cpp:26: unknown type name 'constexpr'
E         + WARNING: namespace.cpp:52: scoped enumerations are a C++11 extension
```

and

```
E         + ERROR: class.cpp:28: unknown type name 'constexpr'
E         + ERROR: class.cpp:34: expected ';' at end of declaration list
E         + ERROR: class.cpp:37: unknown type name 'constexpr'
E         + WARNING: class.cpp:70: deleted function definitions are a C++11 extension
E         + WARNING: class.cpp:73: 'override' keyword is a C++11 extension
E         + WARNING: class.cpp:91: defaulted function definitions are a C++11 extension
E         + /Users/runner/work/github-actions-playground/github-actions-playground/test/cpp/class.cpp:25: WARNING: Error in declarator or parameters-and-qualifiers
E         + If pointer to member declarator:
E         +   Invalid C++ declaration: Expected identifier in nested name. [error at 48]
E         +     public static constexpr  static_constexpr_member
E         +     ------------------------------------------------^
E         + If declarator-id:
E         +   Invalid C++ declaration: Expected identifier in nested name. [error at 48]
E         +     public static constexpr  static_constexpr_member
E         +     ------------------------------------------------^
```

Additionally, there are some failures to find C++ headers. #252 improved the situation in that the paths aren't wrong, but still issues.

jnikula:
Anyway, the pull request at hand is a prerequisite no matter what. #253 has also been very helpful.

BrunoMSantos:
Good to merge I think!

jnikula:
> Good to merge I think!

Thanks, merged!

Can you see the detailed github action results, with logs etc, for my pull requests?

BrunoMSantos:
> Can you see the detailed github action results, with logs etc, for my pull requests?

I've had a quick look yesterday, but I need to take a bit longer on that one. Haven't seen your latest updates at all though.

jnikula:
> I've had a quick look yesterday, but I need to take a bit longer on that one. Haven't seen your latest updates at all though.

Okay, with the prep work merged, I can just send a workflow PR. Mostly I've been playing in a private repo.

jnikula:
@BrunoMSantos Increasingly the wrong place to discuss, but anyway: Looks like we need to specify the C++ standard to use for a lot of tests. The default is different on macos. I just added `-std=c++17` to the `.yaml` files of a bunch of tests. Is that okay, or should we add a default in `testenv.py` that could be overridden in `.yaml`?

Still having include issues, and it's driving me nuts. Xcode libclang can't find C++ headers, but homebrew version can. Xcode libclang can find C headers, but homebrew version can't. If one or the other worked, I'd just call it a day.

BrunoMSantos:
I only have #249 to go by. Care to make a draft PR with the latest version?

> Looks like we need to specify the C++ standard to use for a lot of tests. The default is different on macos. I just added -std=c++17 to the .yaml files of a bunch of tests. Is that okay, or should we add a default in testenv.py that could be overridden in .yaml?

Other than that, assuming a default may be wise (perhaps the difference is in the clang version each system uses by default, in which case it doesn't hurt to be explicit). Adding it in every test case is not so good I think as the test developer may not immediately know he needs to.

> Still having include issues, and it's driving me nuts. Xcode libclang can't find C++ headers, but homebrew version can. Xcode libclang can find C headers, but homebrew version can't. If one or the other worked, I'd just call it a day.

Alas I know nothing of macos, but that's weird indeed. How are you testing this? The CI itself or do you have a system on your end? I.e. can I see the output?

jnikula:
> I only have #249 to go by. Care to make a draft PR with the latest version?

Sure. #254

PR #249, opened 2024-09-01
Macos GitHub workflow

jnikula:
#248 + the actual workflow.

Still fails, but it's a start.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -42,8 +42,8 @@ jobs:
           . venv
           make check-examples
 
-  test:
-    name: Run test suite
+  test-ubuntu:
+    name: Run test suite (Ubuntu)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


jnikula:
Any C++ tests that `#include` something hit this:

```
E         + ERROR: c++/v1/cstddef:46: <cstddef> tried including <stddef.h> but didn't find libc++'s <stddef.h> header.           This usually means that your header search paths are not configured properly.           The header search paths should contain the C++ Standard Library headers before           any C Standard Library, and you are probably using compiler flags that make that           not be the case.
E         + ERROR: c++/v1/cstddef:59: no member named 'nullptr_t' in the global namespace
E         + ERROR: c++/v1/__type_traits/is_null_pointer.h:26: use of undeclared identifier 'nullptr_t'
E         + ERROR: c++/v1/cstdint:149: <cstdint> tried including <stdint.h> but didn't find libc++'s <stdint.h> header.           This usually means that your header search paths are not configured properly.           The header search paths should contain the C++ Standard Library headers before           any C Standard Library, and you are probably using compiler flags that make that           not be the case.
```

despite the attempts to figure out the C++ header search path from the right compiler for the right language.

The error messages were added in https://reviews.llvm.org/D131441 and the implication is that the include paths passed on the command-line are wrong... but then libclang does not figure them out itself, you have to pass them.

PR #250, opened 2024-09-06
doc: remove notes about C++ gaps

jnikula:
While there may still be gaps with the C++ support, we do have the most common cases covered, and don't want to scare away potential users.


BrunoMSantos (approved):
Sounds fair. And if we end up getting more issue reports related to it, all the better. I'm sure there are plenty of sharp corners still.

PR #252, opened 2024-09-08
Use libclang include path in tests

jnikula:
Sort out C header stuff, update Docker images, and stop asking clang for the header search path during testing. This is no longer an issue with the current images.


BrunoMSantos (approved):
This looks great!

Sorry it took me a while to start reviewing these, but I'm back now. Should be going over the following PRs soon as well :)

jnikula:
> This looks great!
> 
> Sorry it took me a while to start reviewing these, but I'm back now. Should be going over the following PRs soon as well :)

Don't worry about it, and thanks for going through them!

PR #253, opened 2024-09-11
test: combine the asserts for output and errors

jnikula:
With the output and error asserts separate, the test case ends on the first failing assertion, and we never know what was in errors. But the clue to the output mismatch might be in the errors. For example, an include file was not found. By combining the asserts into one, we get all the differences in the pytest results.

### DIFF
--- a/test/testenv.py
+++ b/test/testenv.py
@@ -104,8 +104,7 @@ class Testcase:
 
         assert output_docs, 'empty output'
         assert expect_docs, 'empty expected'
-        assert output_docs == expect_docs
-        assert output_errors == expect_errors
+        assert output_docs + output_errors == expect_docs + expect_errors
 
 def get_testid(testcase):
     return testcase.get_testid()


test/testenv.py
```diff
@@ -104,8 +104,7 @@ def run_test(self):
 
         assert output_docs, 'empty output'
         assert expect_docs, 'empty expected'
-        assert output_docs == expect_docs
-        assert output_errors == expect_errors
+        assert output_docs + output_errors == expect_docs + expect_errors
```
BrunoMSantos, line 107: Is it worth to `'\n'.join(...)` to improve the diff or is `output_docs` guaranteed to end in a new line?

BrunoMSantos (approved):
Good idea, I had a minor comment only, but it's good to go either way.

test/testenv.py
jnikula: Per my testing, there's always been a newline, and the diff has looked fine with this. (And I've looked at plenty when trying to debug the macos CI issues.)

Issue #255, opened 2024-09-14
It's not possible to define clang args separately for C and C++ in mixed language projects


jnikula:
If you have a mixed C and C++ language project, and need to pass args to clang, `hawkmoth_clang` is inadequate. Simply specifying the `-std` to use falls apart, because `hawkmoth_clang` gets passed to both C and C++.

The directive `:clang:` option is domain specific by design, but it's not feasible to defaults spanning entire projects.

Adding default clang args to Hawkmoth tests suffers from the same issue.

PR #256, opened 2024-09-14
test: define default clang args for testing

jnikula:
For starters, define the language stardard to use. Use c17 and c++17, fairly arbitrarily. Tests can override this. We can remove the std option from template.yaml.

This is a bit more convoluted than should be necessary, for a few reasons:

- Extension tests use the directive options from directive string. For that, the defaults should be specified in conf.py, but hawkmoth_clang is not domain specific, and can't be used. See [1]. We need to insert the defaults in directive options.

- We want to preserve the original directive options, and present those in examples documentation. However, in all testing, we need to amend the defaults.

Note that the parser selects the -x language option based on domain and filename suffix.

[1] https://github.com/jnikula/hawkmoth/issues/255

### DIFF
--- a/test/cpp/template.yaml
+++ b/test/cpp/template.yaml
@@ -3,7 +3,4 @@ directives:
   directive: autodoc
   arguments:
   - template.cpp
-  options:
-    clang:
-    - --std=c++17
 expected: template.rst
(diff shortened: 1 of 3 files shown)


jnikula:
Got a better idea, dropping this one.

PR #257, opened 2024-09-14
test: rename test_cautodoc.py -> test_extension.py

jnikula:
cautodoc is a remnant from the past. This module tests the extension, name it accordingly.

Also rename the test methods to reflect the same.


BrunoMSantos (approved):
:+1:

PR #258, opened 2024-09-14
Domain specific clang args

### LINKED ISSUE #255: It's not possible to define clang args separately for C and C++ in mixed language projects

jnikula:
Fix #255 and specify C and C++ standard to use for tests.

Importing conf.py in tests requires some changes in update-examples and mypy.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ Unreleased `master`_
 
 :Date: YYYY-MM-DD
 
+Added
+~~~~~
+
+* Domain specific config options ``hawkmoth_clang_c`` and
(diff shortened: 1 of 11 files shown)


BrunoMSantos (approved):
Great one!

PR #259, opened 2024-09-23
Developer documentation

jnikula:
Start a place under doc/ to gather developer documentation. Add contributing guide and test readme for starters.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -83,7 +83,9 @@ html_theme_options = {
     'extra_nav_links': {
         'GitHub': 'https://github.com/jnikula/hawkmoth',
         'PyPI': 'https://pypi.org/project/hawkmoth',
-    }
+    },
+    'sidebar_width': '280px',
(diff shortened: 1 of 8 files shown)


doc/conf.py
```diff
@@ -83,7 +83,9 @@
     'extra_nav_links': {
         'GitHub': 'https://github.com/jnikula/hawkmoth',
         'PyPI': 'https://pypi.org/project/hawkmoth',
-    }
+    },
+    'sidebar_width': '280px',
+    'page_width': '1000px',
```
BrunoMSantos, line 88: Are hidpi screens handled automatically through scaling? I think not, but bear in mind I know very little of CSS.

BrunoMSantos (approved):
Lgtm ;)

doc/conf.py
jnikula: Not my strong suit either, and I can only justify this because the original Alabaster stylesheet uses px.

PR #261, opened 2024-09-28
typing: use Optional[T] instead of T | None for Python 3.9 compatibility

jnikula:
Although Python 3.9 supports built-in type hinting instead of having to import from typing, unfortunately the T | None syntax was only introduced in Python 3.10.

Use Optional[T] instead for Python 3.9 compatibility.


BrunoMSantos (approved):
Too bad :/
Looks good in the meantime though.

PR #265, opened 2025-02-08
Autoconf tweaks

jnikula:
I toyed with the new autoconf stuff and ran into some small issues. Fix and tweak it a bit.


BrunoMSantos:
Shiny. Looks great!

PR #266, opened 2025-02-09
macos xcode CI

jnikula:
Nothing for you to see here. Just a test run.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -91,7 +91,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
-      - run: printenv | sort
+      - name: Environment
+        run: printenv | sort
+      - name: xcode-select
+        run: xcode-select --print-path
+      - name: Files
+        run: find $(xcode-select --print-path)
       - name: Test
         run: |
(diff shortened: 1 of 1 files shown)


jnikula:
So it works. Though I think this is using homebrew clang from PATH, and thus homebrew headers instead of xcode.

I'm not sure if we want to add this to CI though.

BrunoMSantos:
> Though I think this is using homebrew clang from PATH, and thus homebrew headers instead of xcode.

Dunno. Why out of curiosity?

> I'm not sure if we want to add this to CI though.

I know nothing of mac world to judge how relevant it is. That said, we don't test for alternate clang installations in each distro. I'm sure it works for any other one, standard or not, provided paths are sorted / minimum versions met.

jnikula:
> > Though I think this is using homebrew clang from PATH, and thus homebrew headers instead of xcode.
> 
> Dunno. Why out of curiosity?

I think it would be better to set both `DYLD_LIBRARY_PATH` and `PATH` to pick libclang and clang from the same toolchain, like we do for homebrew currently. I just didn't know what to set to `PATH` to use xcode clang. :shrug: 

> > I'm not sure if we want to add this to CI though.
> 
> I know nothing of mac world to judge how relevant it is. That said, we don't test for alternate clang installations in each distro. I'm sure it works for any other one, standard or not, provided paths are sorted / minimum versions met.

Yeah, mostly I just sent this because I had a bunch of branches with old attempts to make this work, so I wanted to see if it works. And it does, my curiosity is satisfied.

I think I'll just close this one.

PR #268, opened 2025-03-29
Use -isystem

jnikula:
I was planning on a release, and ran the test suite on all docker containers again. Fedora Rawhide failed due to warnings in GCC 15 C++ headers. Using `-isystem` instead of `-I` suppresses the warnings.

### DIFF
--- a/src/hawkmoth/util/compiler.py
+++ b/src/hawkmoth/util/compiler.py
@@ -64,9 +64,13 @@ def get_include_args(cpath='clang', lang='c', cc_path=None):
     return ['-nostdinc'] + [f'-isystem{path}' for path in _get_include_paths(cpath, lang)]
 
 if __name__ == '__main__':
+    import argparse
     import pprint
-    import sys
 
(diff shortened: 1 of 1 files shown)


BrunoMSantos (approved):
Looks and tests ok on my end. It's how it was meant to be in fact.

I remember I had an issue with this before, but seems to work now. The flag is still not listed in my man pages, but it is listed in `clang --help` :man_shrugging: 

I don't think there was any (major) clang update since, but did you test in more conservative distros after this patch? Ubuntus and such?

jnikula:
> Looks and tests ok on my end. It's how it was meant to be in fact.
> 
> I remember I had an issue with this before, but seems to work now. The flag is still not listed in my man pages, but it is listed in `clang --help` 🤷‍♂️

Right, I remember this was mentioned.

> I don't think there was any (major) clang update since, but did you test in more conservative distros after this patch? Ubuntus and such?

Yes, I now regularly run the tests on the containers under `docker/`.

Thanks for the review!

PR #271, opened 2025-04-12
Use SPDX license/copyright headers, and go for REUSE compliance

jnikula:
The TL;DR of the REUSE project: "We want to improve the way that people license their software projects such that it is easy, comprehensive, unambiguous, and machine-readable." https://reuse.software/

Switch to SPDX license/copyright headers, and enable REUSE compliance checks via `make check-reuse` and github CI.

@BrunoMSantos The only change that I think could be controversial is switching from copyright year ranges e.g. 2019-2022 to only the year of initial publication e.g. 2019. I think it makes sense (avoids having to change these all the time, but still keeps the year there), but I don't want to do it it you don't want to. See https://reuse.software/faq/#years-copyright and https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-keep-the-year for discussion.

I'm also adding your copyright notices to some files that I think are missing them. Please let me know if that's okay or if I missed some.


BrunoMSantos:
Looks good to me! SPDX is neat, though not a massive improvement over what we had. But the reuse tools will help with consistency, and that's always good.

Side note: I don't care too much about copyright assignments honestly, especially as it's always a bit arbitrary what constitutes copyright worthy changes. I would care that multiple people had copyright attributions if we had a copyleft license, but I already made my pitch for that. And I'd care only to guarantee license immutability. Given how permissive BSD is though, I honestly don't think it makes any tangible difference, it's barely any different from public domain :shrug:

jnikula:
> Looks good to me! SPDX is neat, though not a massive improvement over what we had. But the reuse tools will help with consistency, and that's always good.

Thanks, merged!

> Side note: I don't care too much about copyright assignments honestly, especially as it's always a bit arbitrary what constitutes copyright worthy changes. I would care that multiple people had copyright attributions if we had a copyleft license, but I already made my pitch for that. And I'd care only to guarantee license immutability. Given how permissive BSD is though, I honestly don't think it makes any tangible difference, it's barely any different from public domain 🤷

I think my only reason for going with BSD originally was that Sphinx itself is BSD licensed, and most of the extensions in the sphinx-contrib collection are also BSD licensed. I couldn't figure out what it would really mean to have a GPL-licensed plugin to a BSD-licensed software, or how https://www.gnu.org/licenses/gpl-faq.html#GPLPlugins applies to Python. I'm not sure if I'm any wiser on this now, even if I'm overall more likely to go with GPL than before.

Even now, I don't think relicensing itself would be a problem if we so decided.

BrunoMSantos:
> I think my only reason for going with BSD originally was that Sphinx itself is BSD licensed, and most of the extensions in the sphinx-contrib collection are also BSD licensed. I couldn't figure out what it would really mean to have a GPL-licensed plugin to a BSD-licensed software,

Yeah, I know. Though according to the FSF BSD is [compatible](https://www.gnu.org/licenses/license-list.en.html#ModifiedBSD) with GPL. And if it wasn't, then [this](https://www.gnu.org/licenses/gpl-faq.html#GPLPluginsInNF) would apply I think.

> or how https://www.gnu.org/licenses/gpl-faq.html#GPLPlugins applies to Python. I'm not sure if I'm any wiser on this now, even if I'm overall more likely to go with GPL than before.

This would require well demarcated exceptions I think. Something like 'documented APIs are fair game, but nothing else'.

> Even now, I don't think relicensing itself would be a problem if we so decided.

Yeah, there's only one other person to ask I think. And maybe check it's not a deal breaker for the Mesa guys? But I'm not hung up on this as you know ;)

Issue #274, opened 2025-04-23
Function typedef missing signatures


overcrook:
Hi, I'm looking for the alternative to the cumbersome doxygen+breathe combination, and this project is almost perfect, great work!

But I'm missing the proper documentation for function typedefs. These function typedefs are widely used for defining callbacks.

Example:
```c
/**
 * This is a callback type
 *
 * :param foo: the foo parameter
 * :param bar: the bar parameter
 * :return: Return some code
 */
typedef int (callback_t)(int foo, const char *bar);
```

Actual output:
```rest
# {'line': 1, 'cursor.kind': CursorKind.TYPEDEF_DECL, 'cursor.displayname': 'callback_t', 'cursor.spelling': 'callback_t'}

.. c:type:: callback_t

   This is a callback type

   :param foo: the foo parameter
   :param bar: the bar parameter
   :return: Return some code
```

This output lacks all the necessary information about the type of callback, the expected parameters and the return type.

Expected output
```rest
.. c:type:: int callback_t(int foo, const char *bar)

   This is a callback type

   :param foo: the foo parameter
   :param bar: the bar parameter
   :return: Return some code
```

Note that in this case the function typedef is used to simplify the code, not to hide the actual types.

jnikula:
> Hi, I'm looking for the alternative to the cumbersome doxygen+breathe combination, and this project is almost perfect, great work!

Thanks for the feedback! 🎉 

> But I'm missing the proper documentation for function typedefs. These function typedefs are widely used for defining callbacks.

Yeah, Sphinx allows two styles for typedefs:

```
.. c:type:: typedef-like declaration
.. c:type:: name
```

and we've just opted for the latter. I'm not sure I like the thought of switching everything to the former, but completely agreed function typedefs should follow it.

I'll look into it.

BrunoMSantos:
> I'm not sure I like the thought of switching everything to the former, [...]

I have the same misgivings, for pretty much the same reasons I believe. But wouldn't say it's completely irrelevant.

> [...], but completely agreed function typedefs should follow it.

Agreed, these really are special: one cannot define a function type without a typedef nor define a truly opaque pointer to one without encapsulating it in a structure. We also have another exception on typedefs already, and a less savory one in my opinion, so this is not unprecedented in a sense.

[...]

PR #275, opened 2025-05-01
Function pointer typedef

jnikula:
Handle #274

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -100,14 +100,31 @@ Directive
 
 .. code-block:: rest
 
+   .. c:autodoc:: typedef.c
+
+Output
+~~~~~~
+
+.. c:autodoc:: typedef.c
+
+
+Directive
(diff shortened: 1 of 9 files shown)


BrunoMSantos (approved):
I think I'd still like to see a switch to (un)hide any typedef definition at some point, but I can't argue this is most of the work done and that it is a big improvement. Somehow I was expecting it to be a bit messier to detect which case is which.

Anyway, big :+1: from me.

jnikula:
> I think I'd still like to see a switch to (un)hide any typedef definition at some point, but I can't argue this is most of the work done and that it is a big improvement. Somehow I was expecting it to be a bit messier to detect which case is which.

I'm always a proponent of sane defaults, and I think this one here is a good compromise. That said, I'm not opposed to adding that option later.

> Anyway, big 👍 from me.

Thanks again for the review!

PR #276, opened 2025-05-01
extension: fix parsed file cache lookup with clang args

### LINKED ISSUE #273: No caching when using :clang:
Hi,

First thanks a lot for this plugin, we started using it after giving up on breathe, and it is very promising to document APIs and export them naturally in an already existing sphinx documentation ! Also a lot lighter and faster both to setup and to run.

I have stumbled on an issue, which may be due to me doing something wrong, but I couldn't find information in the documentation.
In our project, we use custom include path for almost every file. When wanting to analyze a file using hawkmoth, clang will complain if it cannot find an included header. Fine, so I used `:clang:` to specify the include path manually. But the, the feature which makes it possible to use functions already parsed does not work anymore.

So this works:

```rst
.. c:autofunction:: func_a
   :file: foo.c

.. c:autofunction:: func_b
```

But if I instead use a file which requires an include path, it fails to find the symbol (`foo.rst:5: WARNING: "c:autofunction:: func_b" does not match documented symbols.`)

```rst
.. c:autofunction:: func_a
   :file: foo.c
   :clang: -Icustompath

.. c:autofunction:: func_b
```

Whereas, this works:

```rst
.. c:autofunction:: func_a
   :file: foo.c
   :clang: -Icustompath

.. c:autofunction:: func_b
   :file: foo.c
   :clang: -Icustompath
```

Is this intended ? Is there a workaround ? Is there a better way of handling per-file custom compiler options (typically -I include path and -D defines) ?

jnikula:
We look up parsed results based on both filename and clang args. But if we haven't specified the filename, there's no point in using clang args to filter results unless it's explicitly specified.

Previously, this wouldn't work:

```
.. c:autofunction:: func_a
   :file: foo.c
   :clang: -Icustompath

.. c:autofunction:: func_b

```
Because the latter doesn't have matching `:clang:`. Stop requiring it.

Fixes #273

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -129,10 +129,19 @@ class _AutoBaseDirective(SphinxDirective):
         return num_matches
 
     def __get_docstrings(self, viewlist):
+        filter_filenames = self._get_filenames()
+        filter_clang_args = [self.__get_clang_args()]
+
+        # If filenames is None, we're relying on a previous directive to have
+        # parsed the file. In that case, only filter by clang arguments if
+        # they're explicitly specified.
+        if filter_filenames is None and 'clang' not in self.options:
+            filter_clang_args = None
+
         num_matches = 0
-        for root in self.__parsed_files(filter_filenames=self._get_filenames(),
+        for root in self.__parsed_files(filter_filenames=filter_filenames,
                                         filter_domains=[self._domain],
-                                        filter_clang_args=[self.__get_clang_args()]):
+                                        filter_clang_args=filter_clang_args):
             num_matches += self.__get_docstrings_for_root(viewlist, root)
 
         if num_matches == 0:
--- a/test/c/caching.yaml
+++ b/test/c/caching.yaml
@@ -17,5 +17,4 @@ directives:
   - foo_struct
   options:
     members:
-expected-failure: true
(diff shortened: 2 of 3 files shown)


test/c/autostruct.yaml
```diff
@@ -5,6 +5,8 @@ directives:
   - sample_struct
   options:
     file: struct.c
+    clang:
+    - -DTEST_CACHE_MATCHING_WITH_CLANG_ARGS
```
BrunoMSantos, line 9: Spurious / incomplete change?

BrunoMSantos (approved):
Looks good apart from the unit test change ;)

test/c/autostruct.yaml
BrunoMSantos: Oh... Took me a bit.

Yeah, this would probably benefit from a commit message of its own if not a comment in the yaml / separate unit test. This is effectively part of a *different unit test* than the one it changes. Am I reading it right?

BrunoMSantos:
Yup, this so much clearer. Hope it wasn't too bad :)

jnikula:
> Yup, this so much clearer. Hope it wasn't too bad :)

Not at all, and it was the right thing to do. Thanks for the review!

PR #278, opened 2025-06-12
Use the special function typedef form when using the `autotype` directive

skinnynerd:
Previously, using the `autodoc` directive would recognize function pointer typedefs and show their arguments.  However, using the `autotype` directive would not pick up on this and wouldn't show the type.  This merge request adds the Typedef Function Docstring types to the allowed docstrings for the `autotype` directive.


skinnynerd:
Previously, the test I added would fail with "does not match any documented symbols" because function typedefs would not be considered for inclusion in the `autotype` directive.  That meant there was no directive which could individually document a function typedef.  The test would work properly if the `autodoc` directive was used, since that includes all docstrings regardless of type.

I updated the commit message.  Please let me know if there's any other information you need.

jnikula:
> Previously, the test I added would fail with "does not match any documented symbols" because function typedefs would not be considered for inclusion in the `autotype` directive. That meant there was no directive which could individually document a function typedef. The test would work properly if the `autodoc` directive was used, since that includes all docstrings regardless of type.

Ah! This actually worked before, and it was #275 that broke it when adding the dedicated function pointer typedefs!

Thanks for the fix, and the improved test coverage, appreciated!

> I updated the commit message. Please let me know if there's any other information you need.

Thanks. I took the liberty of tweaking the commit message some more to point at the regressing commit, and pushed manually.

PR #279, opened 2025-06-12
Use :any: instead of :ref: for Javadoc cross-references

skinnynerd:
Javadoc parsing would replace Javadoc `@ref` with RST `:ref:`.  This only points to labels, not C directives.  A decent alternative is to just use `:any:` to point to anything we can find with the right name.

### DIFF
--- a/src/hawkmoth/ext/javadoc/__init__.py
+++ b/src/hawkmoth/ext/javadoc/__init__.py
@@ -47,7 +47,7 @@ class _handler:
 
         # references to previous anchors
         # FIXME: link title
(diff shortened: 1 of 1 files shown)


jnikula:
Sounds reasonable.

Please describe what's being done in the commit message too. I maintain a rebase-and-merge style here, so the pull request descriptions aren't recorded to git, only github.

skinnynerd:
Thanks for the review! Updated the commit message to match the pull request description.  Please let me know if there's more information you would like.

Issue #281, opened 2025-06-16
Add example for "hawkmoth-process-docstring"


dvdhrm:
Hi

With the deprecation of `cautodoc_transformations` in #246 there is no example of how to apply simple transformations in downstream users. I would love to add an example, but I thought we can maybe first discuss whether you are willing to include such examples, and how it should look like.

We have been using `cautodoc_transformations` to make use of hawkmoth with our own kerneldoc'ish comments, and currently look into getting rid of the deprecated features. My current approach is to simply copy `src/hawkmoth/ext/transformations/__init__.py` next to `conf.py`, and then include it via `extensions`. I then adjust the module to inline my transformations and apply them directly.

Unfortunately, this requires putting the transformations into a separate module, instead of inlineing it in conf.py as before. This also requires putting `./conf.py` into `PYTHONPATH`, since otherwise `sphinx-build ...` will not find it when used with relative paths.

Is there any better option to apply ad-hoc transformations in downstream users?

Thanks
David

jnikula:
Hi @dvdhrm, thanks for the report!

Adding an example in the documentation is certainly a good idea, especially for the small-ish ad-hoc transformations. I guess it's not common knowledge that `conf.py` can itself be an extension! See https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-extensions

It should suffice to add a `setup()` function in `conf.py`, and connect the `hawkmoth-process-docstring` event there. There's no need for a separate module and import, and this should make it much easier to do the ad-hoc stuff.

I'll expand on this when I have a bit more time, but maybe that's enough to get you going.

PS. The reason we deprecated `cautodoc_transformations` was that it leads to the configuration having object/function references, they don't cache nicely on disk, and it leads to unnecessary rebuilds even when there's nothing to rebuild.

jnikula:
The Hawkmoth `conf.py` is an extension: https://github.com/jnikula/hawkmoth/blob/master/doc/conf.py#L126

dvdhrm:
Ah, this is great. This is exactly what I was looking for! Thanks a lot! I will see if I can come up with a small example.

PR #282, opened 2025-06-17
doc/tips: describe ad-hoc extensions

dvdhrm:
Add a section to describe how Sphinx supports ad-hoc extensions. That is, you can just define extensions from within your `conf.py` configuration.


dvdhrm:
(Fixes #281)

PR #286, opened 2025-08-17
doc/tips: describe ad-hoc extensions

jnikula:
Add a section to describe how Sphinx supports ad-hoc extensions. That is, you can just define extensions from within your `conf.py` configuration.

[Jani: Minor stylistic changes, improved the example slightly.]

Basically v2 of #282 by @dvdhrm, to get that and #281 closed.


BrunoMSantos:
Great idea, I can see how this massively improves the experience for newcomers ;)

The patch looks good by the way, in case you were waiting for input.

PR #287, opened 2025-08-21
Refactor docstring filtering

jnikula:
The filtering in `Docstring.walk()` has always been cumbersome at best, as described by the comment:

```
        # Note: The filtering is pretty specialized for our use case here. It
        # only filters the immediate children, not this comment, nor
        # grandchildren.
```

Refactor all of the filtering out of `Docstring` to the calling side, which has a much better idea what and how to filter. The division of responsibilities is better, and `Docstring` becomes more clear.

Initially, the filtering on the calling side (the extension and parser test) is 1:1 to the original, and further cleanups are left for the future. Also, this should make fancier filtering such as  #284 easier to implement cleanly.

### DIFF
--- a/test/test_parser.py
+++ b/test/test_parser.py
@@ -97,6 +97,13 @@ class ParserTestcase(testenv.Testcase):
             if filter_filenames is None and 'clang' not in directive.options:
                 filter_clang_args = None
 
+            def get_docstring(ds):
+                transform = directive.options.get('transform')
+                def process_docstring(lines): return _process_docstring(transform, lines)
+
+                lines, _ = ds.get_docstring(process_docstring=process_docstring)
+                return '\n'.join(lines) + '\n'
+
             for root in roots.values():
                 if filter_filenames is not None and root.get_filename() not in filter_filenames:
                     continue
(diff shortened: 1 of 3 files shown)


test/test_parser.py
```diff
@@ -114,9 +114,14 @@ def process_docstring(lines): return _process_docstring(transform, lines)
                 if filter_clang_args is not None and root.get_clang_args() not in filter_clang_args:
                     continue
 
-                for docstrings in root.walk(recurse=False, filter_types=_filter_types(directive),
-                                            filter_names=_filter_names(directive)):
-                    for docstr in docstrings.walk(filter_names=_filter_members(directive)):
+                for primary in root:
+                    if _filter_types(directive) is not None and type(primary) not in _filter_types(directive):  # noqa: E501
```
BrunoMSantos, line 118: Nitpick: some of these lines are already mouthfuls, but these ones require an exception too. I may be biased though, I'm a stickler for 79 character limit :P

There's at least one more commit with a similar `E501` if you care to fix it.

BrunoMSantos (approved):
Sorry again for the delay, and thanks for waiting. It's a good clean up and have nothing to add, but it's an area I want to keep tabs on if possible. I do have some plans here if I ever find the time :)

test/test_parser.py
jnikula: Thanks a lot for the review! I've got a bunch of follow-up changes lined up, including cleaning these up. So I think I'll just merge this.

I usually like to keep the lines under about a 100 chars or so, and I think that's what I've set up for flake8 config here. I think 80 is too restrictive. ;)

PR #288, opened 2025-08-30
Refactor docstring processing

jnikula:
Follow-up to #287 with a bunch more refactoring.

The primary change is adding a `DocstringProcessor` class for, well, processing docstrings, instead of passing in functions that require lambdas in the callers.

While no new Sphinx events are added here, this makes it much easier to add events that you could pass the `DocCursor` and `header_lines` to.

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -110,13 +110,17 @@ class _AutoBaseDirective(SphinxDirective):
             viewlist.append(line, root.get_filename(), line_number - 1)
             line_number += 1
 
+    @staticmethod
+    def _skip(thing, iterable):
+        return iterable is not None and thing not in iterable
+
     def __get_docstrings_for_root(self, viewlist, root):
         num_matches = 0
         for primary in root:
-            if self._docstring_types is not None and type(primary) not in self._docstring_types:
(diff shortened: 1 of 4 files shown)


jnikula:
PS. I'm not really sure if `DocstringProcessor` is a great name or whether it shoud be placed in `docstring.py` or somewhere else. But I guess none of this is cast in stone, they can be renamed and relocated later as needed.

src/hawkmoth/__init__.py
```diff
@@ -155,13 +155,13 @@ def __get_docstrings(self, viewlist):
 
         num_matches = 0
         for root in self.__parsed_files():
-            if filter_filenames is not None and root.get_filename() not in filter_filenames:
+            if self._skip(root.get_filename(), filter_filenames):
```
BrunoMSantos, line 158: Hmm, I find the imbalance a bit triggering. The other ones gained specific `skip_something` methods, these ones use the `_skip` one.

I get why you did it this way here, but it's still a bit of cognitive load spent understanding the difference. I'd be tempted at least to remove the other `skip_*` ones and just call `_skip` everywhere.

BrunoMSantos (approved):
Looks great ;)

PR #290, opened 2025-08-30
github: try macos-latest

jnikula:
See if using llvm@18 works.

If it does, I wonder if we should add a matrix for running on both macos-14 and macos-15, or just stick to -latest.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -82,21 +82,21 @@ jobs:
 
   test-macos-homebrew:
     name: Run test suite (macOS homebrew)
-    runs-on: macos-14
+    runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
       - name: Environment
         run: |
-          brew --prefix llvm@15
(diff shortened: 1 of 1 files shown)


jnikula:
@BrunoMSantos Do you have any opinions on this? Or don't care?

There'd be two basic approaches:

- Use `*-latest` images, let them potentially break on image upgrades, and fix them then.
- Use (a matrix of) specific versions of the images, let them go old because who's going to remember to check for upgrades, and fix them then.

BrunoMSantos:
> There'd be two basic approaches:
> 
>     * Use `*-latest` images, let them potentially break on image upgrades, and fix them then.
> 
>     * Use (a matrix of) specific versions of the images, let them go old because who's going to remember to check for upgrades, and fix them then.

If I understand correctly, the issue here was that a specific clang version became unavailable in the newer Apple OS (short of compiling it I suppose). That is no different than us not checking Hawkmoth against all versions of clang in Linux. I.e. latest is good?

If the relevant difference under test was the OS itself somehow, then yeah, we might want to test multiple versions.

PR #300, opened 2026-01-16
extension: fix env.app deprecation warning

jnikula:
Switch from env.app.emit() to env.events.emit() to fix the Sphinx deprecation warning introduced in version v9.0.0:

RemovedInSphinx11Warning: 'sphinx.environment.BuildEnvironment.app' is deprecated

The env.events.emit() interface has been there since Sphinx v3.0.0, so there shouldn't be any need to bump dependencies for this.

See also https://github.com/sphinx-doc/sphinx/issues/14119

### DIFF
--- a/src/hawkmoth/__init__.py
+++ b/src/hawkmoth/__init__.py
@@ -100,7 +100,7 @@ class _AutoBaseDirective(SphinxDirective, docstring.DocstringProcessor):
     def process_docstring(self, lines):
         transform = self.options.get('transform', self.env.config.hawkmoth_transform_default)
 
-        self.env.app.emit('hawkmoth-process-docstring', lines, transform, self.options)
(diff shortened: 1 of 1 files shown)


BrunoMSantos (approved):
Nice to see these recent clean ups in sphinx, even if they break a couple of things along the way.

Just to confirm, this is still valid on older Sphinx versions, right?

BrunoMSantos:
> It should work with Sphinx v3.0.0 and later, based on looking at Sphinx git history, but I was only able to test v6.2.1. Earlier Sphinx would have required an earlier Python version, and I didn't want to go through that rabbit hole.

Ah, sure. I only meant to check that `self.env.events` was not a new API that was missing in earlier versions. I think that's far enough ;)

